PR #15, opened 2020-02-19
attrs -> dataclass

ojii:
There's no real benefit from using the 3rd party attrs over the built in dataclasses, as we don't use its advanced features. The upside is that dataclasses defaults to `auto_attribs=True` (attrs option), which I consider a plus.


ojii:
included in #17

PR #17, opened 2020-03-05
client without (aio)boto(3|core),  2x speedup on query.

ojii:
Not really ready for review yet. 

The goal is to make aiodynamo faster (especially query/scan), since profiling showed that major slowdowns come from botocore itself.

* [x] Properly handle token expiration
* [x] request signing
* [x] request building
* [x] credentials
  * [x] test ec2
  * [x] test ecs
  * [x] test env
* [x] remove boto dependency
  * [x] http interface
  * [x] error handling
  * [x] query builder
* [x] remove boto based client, move "fast" client to top level
* [x] Docs
* [x] fix credentials fetching anti-thundering-herd-system

Maybe:

* [x] better `limit` support


pyproject.toml
```diff
@@ -15,10 +15,10 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.7"
-attrs = ">=18.2"
 aiobotocore= ">=0.10.0"
 boto3 = ">=1.9.49"
-aiohttp = ">=3.5.4"
+httpx = {version = "^0.11.1", extras = ["httpx"]}
```
dimaqq: Does aws support parallel requests over HTTP/2? 😮

ojii: httpx is used for h1 in this case. I just saw no reason to force aiohttp usage so you can bring your own http client and i added  built-in support for httpx and aiohttp.

dimaqq: if it's bring-your-own, then, perhaps, these should be dev-deps, rather than deps? 🤔

src/aiodynamo/fast/client.py
```diff
@@ -0,0 +1,613 @@
...
+            payload=payload,
+            action=action,
+            region=self.region,
+            endpoint=self.endpoint,
+        )
+        return await self.http.post(
+            url=request.url, headers=request.headers, body=request.body
+        )
+
+    async def _fast_depaginate(self, action, payload) -> AsyncIterator[Dict[str, Any]]:
```
dimaqq: it's a bit ugly that payload is changed in place, but I maybe it's ok for an internal function...
as long as there no reties and every caller creates a new payload each time, it's safe, right?

src/aiodynamo/fast/client.py
```diff
@@ -0,0 +1,613 @@
...
+                    # TODO: Limit!
+                    payload["ExclusiveStartKey"] = result["LastEvaluatedKey"]
+                    task = asyncio.create_task(
+                        self.send_request(action=action, payload=payload)
+                    )
+                except KeyError:
+                    task = None
+                yield result
+        except asyncio.CancelledError:
+            if task:
```
dimaqq: I think it should be `if task and not task.done(): task.cancel()`

dimaqq: Ha! it turns out cancelling finished tasks is a-ok, I take my comment back 😓

src/aiodynamo/fast/client.py
```diff
@@ -0,0 +1,613 @@
...
+        )
+
+    async def _fast_depaginate(self, action, payload) -> AsyncIterator[Dict[str, Any]]:
+        task = asyncio.create_task(self.send_request(action=action, payload=payload))
+        try:
+            while task:
+                result = await task
+                try:
+                    # TODO: Limit!
+                    payload["ExclusiveStartKey"] = result["LastEvaluatedKey"]
```
dimaqq: perhaps this is a good place to do:
```
payload = {**payload, "ExclusiveStartKey": result["..."]}
```

src/aiodynamo/fast/http/httpx.py
```diff
@@ -0,0 +1,30 @@
...
+from httpx import AsyncClient
+from yarl import URL
+
+from ..errors import exception_from_response
+from .base import HTTP, Headers, RequestFailed
+
+
+@dataclass(frozen=True)
+class HTTPX(HTTP):
+    client: AsyncClient
```
dimaqq: I'm guessing it would create an `AsyncClient()` if none is passed right?
It would be nice to add a small example of advanced use, e.g. how to pass connection pool, etc, parameters.

src/aiodynamo/fast/credentials.py
```diff
@@ -0,0 +1,306 @@
...
+
+class Disabled(Exception):
+    pass
+
+
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
```
dimaqq: I can't grok what this does or why it's needed 😓

pyproject.toml
ojii: they're "extra" deps. you either install `aiodynamo[httpx]` or `aiodynamo[aiohttp]` or bring your own and install `aiodynamo`

src/aiodynamo/fast/client.py
ojii: that was the idea. it's at the very end and prefixed by `_` with a reason. Doing a `deepcopy` feels bad in a "fast" API.

pyproject.toml
dimaqq: Ohhh that's how it works! :orz:

src/aiodynamo/fast/http/httpx.py
ojii: You have to pass your own. The reason for this is that clients needs to be closed (usually), so that's the callers responsibility (not aiodynamo).

src/aiodynamo/fast/client.py
dimaqq: there's no need for **deep** copy though :)
anyway, I reckon it's acceptable, let's leave a gotcha for some poor future programmer :P

src/aiodynamo/fast/credentials.py
ojii: useful in lambdas (below). I'd prefer calling it `map` but that's already taken in Python.

src/aiodynamo/fast/http/httpx.py
ojii: example how to use it: https://github.com/hennge/aiodynamo/blob/e5fcb911a75431aae51e3b5080ba738b2f7a1646/tests/integration/test_client.py#L85-L87

ojii:
query benchmarks fetch ~35k items from a table running on a t3.micro on EC2:

```
Mean +- std dev: [boto3] 7.58 sec +- 0.27 sec -> [botocore] 6.61 sec +- 0.28 sec: 1.15x faster (-13%)
Mean +- std dev: [boto3] 7.58 sec +- 0.27 sec -> [aiobotocore] 6.68 sec +- 0.27 sec: 1.13x faster (-12%)
Mean +- std dev: [boto3] 7.58 sec +- 0.27 sec -> [aiodynamo_aiohttp] 3.65 sec +- 0.17 sec: 2.08x faster (-52%)
Mean +- std dev: [boto3] 7.58 sec +- 0.27 sec -> [aiodynamo_httpx] 3.80 sec +- 0.24 sec: 2.00x faster (-50%)
```

TL;DR: `FastClient` is about a 2x speedup in this case. As with all benchmarks, your mileage may vary.

ojii:
@stupoid if you could also have a look at this.

benchmarks/query/README.md
```diff
@@ -0,0 +1,4 @@
+Run:
+
+* `BENCH_TABLE_NAME=<table> BENCH_KEY_FIELD=<key-field> BENCH_KEY_VALUE=<key-value> BENCH_REGION_NAME=<region-name> poetry run  python bench_<aiobotocore|aiodynamo_aiohttp|aiodynamo_httpx|boto3|botocore>.py -o <aiobotocore|aiodynamo_aiohttp|aiodynamo_httpx|boto3|botocore>.json --rigorous --inherit-environ BENCH_TABLE_NAME,BENCH_KEY_FIELD,BENCH_KEY_VALUE,BENCH_REGION_NAME,AWS_ACCESS_KEY_ID,AWS_SECRET_ACCESS_KEY,AWS_SESSION_TOKEN`
```
dimaqq: this is unparseable, neither as raw text nor rendered...
Please rephrase, or make it a multiline code sample or something :) 🙏

benchmarks/query/bench_aiobotocore.py
```diff
@@ -0,0 +1,36 @@
+import asyncio
```
dimaqq: Hmmm a module docstring would be nice :)
Should it return something?
The way the timing information is passed... somewhere... is just so odd 😅

benchmarks/query/bench_aiodynamo_aiohttp.py
```diff
@@ -0,0 +1,26 @@
...
+            item
+            async for item in client.query(TABLE_NAME, Key(KEY_FIELD).eq(KEY_VALUE))
+        ]
+
+
+def query_aiodynamo_aiohttp():
+    asyncio.run(inner())
+
+
+Runner().bench_func("query", query_aiodynamo_aiohttp)
```
dimaqq: I'm thinking if smth like `if __name__ == "__main__":` should be here... 
I guess it should work for `runpy.run_module()`?

src/aiodynamo/client.py
```diff
@@ -1,70 +1,76 @@
...
     KeySpec,
     KeyType,
     LocalSecondaryIndex,
-    ProjectionExpr,
     ReturnValues,
     Select,
     StreamSpecification,
     TableDescription,
     TableStatus,
+    ThrottleConfig,
```
dimaqq: I wonder if perhaps these dynamo things should be re-exported at the top-level,
so that users of the library could do `xx = aiodynamo.ThrottleConfig(...)` 🤔

src/aiodynamo/client.py
```diff
@@ -1,70 +1,76 @@
...
+from .expressions import (
+    Condition,
+    KeyCondition,
+    Parameters,
+    ProjectionExpression,
     UpdateExpression,
-    get_projection,
 )
+from .http.base import HTTP
+from .sign import signed_dynamo_request
```
dimaqq: Hmmm... the AWS signing page extolls the virtues of signing the requests when sent over HTTP...
I wonder if signing is required over HTTPS though...
There's https://forums.aws.amazon.com/message.jspa?messageID=705800 which mentions how to send unsigned payloads correctly, but I've no clue if dynamo would accept such payloads... 🤔

src/aiodynamo/sign.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def derive_signing_key(key: Key, instant: Instant, region: str) -> bytes:
+    tmp_key1 = sha256_hmac(("AWS4" + key.secret).encode("utf-8"), instant.date)
+    tmp_key2 = sha256_hmac(tmp_key1, region)
+    tmp_key3 = sha256_hmac(tmp_key2, SERVICE)
+    return sha256_hmac(tmp_key3, "aws4_request")
+
+
+def signed_dynamo_request(
```
dimaqq: Would be nice to benchmark this, I'm sure it's way faster than custom-weird-json-encode-decode, but it would be nice to have some idea what share of the runtime accounts for signing and possible unsigning...

src/aiodynamo/client.py
```diff
@@ -417,67 +497,56 @@ def table(self, name: TableName) -> Table:
...
+            payload["Select"] = select.value
+
+        payload["ExpressionAttributeNames"] = params.get_expression_names()
+        payload["ExpressionAttributeValues"] = params.get_expression_values()
+
+        got = 0
+
+        async for result in self._fast_depaginate("Query", payload):
+            for item in result["Items"]:
+                yield dy2py(item, self.numeric_type)
```
dimaqq: I wonder if it pays to inline `dy2py` here 🤔

src/aiodynamo/client.py
```diff
@@ -417,67 +497,56 @@ def table(self, name: TableName) -> Table:
...
+                if limit:
+                    got += 1
+                    if got >= limit:
+                        return
```
dimaqq: might as well preventively slice `result["Items"]` up to the limit.
somehow the code like `for ...: x += 1` just nags me.

src/aiodynamo/client.py
```diff
@@ -417,67 +497,56 @@ def table(self, name: TableName) -> Table:
...
+        if select:
+            payload["Select"] = select.value
+
+        payload["ExpressionAttributeNames"] = params.get_expression_names()
+        payload["ExpressionAttributeValues"] = params.get_expression_values()
+
+        got = 0
+
+        async for result in self._fast_depaginate("Query", payload):
+            for item in result["Items"]:
```
dimaqq: I'm curious what the overhead of asynchronous iteration might be... 
pages are up to 1MB, and there's network overhead anyway, so async-per-page is warranted.
otoh, items are kinda small, so async-per-item may turn out to be expensive.
Can this be benchmarked? 🙏

src/aiodynamo/client.py
```diff
@@ -315,61 +396,68 @@ def table(self, name: TableName) -> Table:
...
-                ReturnValues=return_values.value,
-                ConditionExpression=condition_expression,
-                ExpressionAttributeNames=expression_attribute_names,
-                ExpressionAttributeValues=expression_attribute_values,
-            )
-        )
+            params = Parameters()
+            payload["ConditionExpression"] = condition.encode(params)
+            payload["ExpressionAttributeNames"] = params.get_expression_names()
+            payload["ExpressionAttributeValues"] = params.get_expression_values()
```
dimaqq: I wonder if there should be a static member like `Parameters.blank_expression_values` 🤔

src/aiodynamo/expressions.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+class Parameters:
+    def __init__(self):
+        self.names: Dict[str, str] = {}
+        self.values: Dict[str, Dict[str, Any]] = {}
+        self.names_gen: Iterator[int] = count()
+        self.values_gen: Iterator[int] = count()
+        self.names_cache: _ParametersCache = {}
+        self.values_cache: _ParametersCache = {}
```
dimaqq: Oh, I was so confused as to why `Parameters` were always created without arguments...
Now I see that it's stateful object and that it's passed around with the idea that information is recorded in it... 
I don't know how this could be done better, but it's confusing 🤔

tests/unit/test_expressions.py
```diff
@@ -0,0 +1,50 @@
...
+        (F("foo") & F("bar"), "#n0,#n1", {"#n0": "foo", "#n1": "bar"}),
+        (F("foo", 0, "bar") & F("bar"), "#n0[0].#n1,#n1", {"#n0": "foo", "#n1": "bar"}),
+        (
+            F("foo", "12", "bar") & F("bar"),
+            "#n0.#n1.#n2,#n2",
+            {"#n0": "foo", "#n1": "12", "#n2": "bar"},
+        ),
+    ],
+)
+def test_project(pe, expression, names):
```
dimaqq: what's `pe`? parameter expression?

tests/unit/test_expressions.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.mark.parametrize(
+    "pe,expression,names",
+    [
+        (F("foo") & F("bar"), "#n0,#n1", {"#n0": "foo", "#n1": "bar"}),
+        (F("foo", 0, "bar") & F("bar"), "#n0[0].#n1,#n1", {"#n0": "foo", "#n1": "bar"}),
+        (
+            F("foo", "12", "bar") & F("bar"),
+            "#n0.#n1.#n2,#n2",
+            {"#n0": "foo", "#n1": "12", "#n2": "bar"},
```
dimaqq: This is kinda hard to read... I guess it's like `a+b=c`? or more like `a→b,c`?

src/aiodynamo/client.py
```diff
@@ -486,94 +555,67 @@ def table(self, name: TableName) -> Table:
...
-                ExpressionAttributeNames=expression_attribute_names,
-                ExpressionAttributeValues=py2dy(expression_attribute_values),
-                Select=Select.count.value,
-            )
-        )
+            payload["FilterExpression"] = filter_expression.encode(params)
+        if index:
+            payload["IndexName"] = index
+        payload["ExpressionAttributeNames"] = params.get_expression_names()
+        payload["ExpressionAttributeValues"] = params.get_expression_values()
```
dimaqq: is it valid to include these elements if key condition and filter expression are both empty?

src/aiodynamo/client.py
```diff
@@ -583,38 +625,79 @@ def table(self, name: TableName) -> Table:
...
-                ExpressionAttributeNames=expression_attribute_names,
-                ExpressionAttributeValues=py2dy(expression_attribute_values),
-                ConditionExpression=condition_expression,
-                ReturnValues=return_values.value,
-            )
-        )
+            payload["ConditionExpression"] = condition.encode(params)
+
+        payload["ExpressionAttributeNames"] = params.get_expression_names()
+        payload["ExpressionAttributeValues"] = params.get_expression_values()
```
dimaqq: likewise, is it possible to update expression to be essentially empty, and in that case, is it correct to add attribute names/values fields?

src/aiodynamo/expressions.py
```diff
@@ -0,0 +1,537 @@
...
+    def encode(self, params: Parameters) -> str:
+        pass
+
+
+@dataclass(frozen=True)
+class NotCondition(Condition):
+    base: Condition
+
+    def encode(self, params: Parameters) -> str:
+        return f"NOT {self.base.encode(params)}"
```
dimaqq: Is it OK without parentheses?
As is does dynamo accept smth like `x OR NOT y` or does it demand `x OR (NOT y)` ?

src/aiodynamo/expressions.py
```diff
@@ -0,0 +1,537 @@
...
+        return f"({self.lhs.encode(params)} AND {self.rhs.encode(params)})"
+
+
+@dataclass(frozen=True)
+class OrCondition(Condition):
+    lhs: Condition
+    rhs: Condition
+
+    def encode(self, params: Parameters) -> str:
+        return f"({self.lhs.encode(params)} OR {self.rhs.encode(params)})"
```
dimaqq: I have some vague recollection that dynamo had weird requirements on what must be and what must not be parenthesized... Consider adding tests (against real dynamo) for cases like:

`x&y&z` if that should be `x AND y AND z` vs. `(x AND y) AND z` vs. `x AND (y AND z)`...
`x&y|z`, `x|y&z`, `x&!z`, etc..

src/aiodynamo/expressions.py
```diff
@@ -0,0 +1,537 @@
...
+            encoded
+            for encoded in (params.encode_value(value) for value in self.values)
+            if encoded is not None
+        ]
+        if len(encoded_values) < 1:
+            raise ValueError("IN Condition requires at least one value")
+        if len(encoded_values) > 100:
+            raise ValueError("IN Condition may not contain more than 100 values")
+        values = ",".join(encoded_values)
+        return f"{params.encode_path(self.field.path)} IN ({values})"
```
dimaqq: not sure about parentheses here too... things like `a & x in (vvv) & b`

src/aiodynamo/client.py
```diff
@@ -220,45 +234,64 @@ def table(self, name: TableName) -> Table:
...
+            attempts = 0
+            while attempts < wait_for_active.max_attempts:
+                try:
+                    description = await self.describe_table(name)
+                    if description.status == TableStatus.active:
+                        return
+                except TableNotFound:
+                    pass
+                attempts += 1
+                await asyncio.sleep(wait_for_active.retry_delay)
```
dimaqq: I wonder if an async retry iterator (helper) could be implemented in the `WaitConfig`...
I mean if it can be implemented in a way that's not too ugly... 🤔

src/aiodynamo/client.py
```diff
@@ -583,38 +625,79 @@ def table(self, name: TableName) -> Table:
...
+            try:
+                return await self.http.post(
+                    url=request.url, headers=request.headers, body=request.body
+                )
+            except Throttled:
+                await asyncio.sleep(self.throttle_config.delay_func(attempt))
+                attempt += 1
+        raise Throttled()
+
+    async def _fast_depaginate(
```
dimaqq: it seems the only depaginate that's left is `_fast_depaginate`... could it be renamed to `_depaginate` then?

src/aiodynamo/credentials.py
```diff
@@ -0,0 +1,312 @@
...
+        self._metadata: Optional[Metadata] = None
+
+    async def get_key(self, http: HTTP) -> Optional[Key]:
+        try:
+            await self._check_metadata(http)
+        except Disabled:
+            return None
+        return self._metadata and self._metadata.key
+
+    @highlander
```
dimaqq: how does `highlander` work with multiple regions?
is there still only 1 metadata, or does it accidentally interlock credentials between regions?
(I've no idea at all how mutli-region-ness even works, maybe @michaelaw320 can elaborate)

src/aiodynamo/client.py
```diff
@@ -583,38 +625,79 @@ def table(self, name: TableName) -> Table:
...
+        """
+        Internal API to depaginate the results from query/scan/count.
+        Don't call this directly, use .query, .scan or .count instead.
+        """
+        task = asyncio.create_task(self.send_request(action=action, payload=payload))
+        try:
+            while task:
+                result = await task
+                try:
+                    # TODO: Limit!
```
dimaqq: is the comment perhaps no longer relevant because limit is enforced by higher level?

src/aiodynamo/errors.py
```diff
@@ -1,10 +1,165 @@
...
+class ItemNotFound(AIODynamoError):
     pass
 
 
-class TableNotFound(Exception):
+class UnknownError(AIODynamoError):
+    def __init__(self, status: int, body: bytes):
+        self.status = status
+        self.body = body
+        super().__init__(body)
```
dimaqq: It would be nice to offer custom `__str__` and/or `__repr__` for this exception.

src/aiodynamo/errors.py
```diff
@@ -1,10 +1,165 @@
...
+}
+
+
+def exception_from_response(status: int, body: bytes) -> Exception:
+    if status == 500:
+        return InternalDynamoError()
+    try:
+        data = json.loads(body)
+        return ERRORS[data["__type"].split("#", 1)[1]](data)
+    except:
```
dimaqq: Either `except Exception:`,
or `except (KeyError, ValueError, WhateverOtherError):`

src/aiodynamo/errors.py
dimaqq: Eh, now I see that it inherits stringy methods from Exception, so `body` will be reported, but not status...
Maybe that's enough 🤔

src/aiodynamo/expressions.py
```diff
@@ -0,0 +1,537 @@
...
+        cache: _ParametersCache,
+        cache_key: Any,
+    ) -> str:
+        try:
+            return cache[cache_key]
+
+        except KeyError:
+            can_cache = True
+        except TypeError:
+            can_cache = False
```
dimaqq: odd blank line... I'd rather expect a  blank line after `try: xx; except; except` block...

src/aiodynamo/client.py
ojii: that's for S3 which uses very different authentication strategies. DynamoDB requires all requests to be signed.

benchmarks/query/bench_aiobotocore.py
ojii: i've updated the readme quite a bit, hope that helps.

src/aiodynamo/client.py
ojii: I like having the namespaces.

src/aiodynamo/client.py
ojii: it's ugly, I agreee. but there's multiple `result['Items']`, so slicing wouldn't be that straight forward.

Ideally `_fast_depaginate` would deal with limits but then I'd have to have a separate one for `count` or the API of `_fast_depaginate` becomes very convoluted.

src/aiodynamo/client.py
ojii: I do not understand what your concern is here.

src/aiodynamo/client.py
ojii: I don't understand what you mean.

src/aiodynamo/expressions.py
ojii: It's an internal API and users should never use it directly, so I think it's okay if it's not the best API. To prevent name clashes, all names and values get "placeholderified" by aiodynamo and this keeps track of all of them per API call. It's not the prettiest, I agree, but it get's the job done...

tests/unit/test_expressions.py
ojii: Projection Expression.

tests/unit/test_expressions.py
ojii: which part is hard to read? I don't know what a, b and c are.

src/aiodynamo/client.py
ojii: `count` requires a Key Condition and Key Conditions are not allowed to be empty.

src/aiodynamo/client.py
ojii: if Update Expression is empty, it raises an exception in line 636.

src/aiodynamo/expressions.py
ojii: according to the [docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html#Expressions.OperatorsAndFunctions.Precedence), you can parenthesize any condition. [boto3](https://github.com/boto/boto3/blob/develop/boto3/dynamodb/conditions.py#L224) also does the parens like this for AND/OR

src/aiodynamo/expressions.py
ojii: `x in(y)` is a single thing in dynamo so no parens around it needed.

src/aiodynamo/client.py
ojii: I tried a generic `wait` implementation, but the exception handling etc got very ugly very quick, so I resorted to this solution. I'm not happy with it but can't come up with a sane API.

src/aiodynamo/client.py
ojii: but. it's fast!

src/aiodynamo/credentials.py
ojii: The client itself is region-locked (you explicitly pass it a region), so that's not a problem. The highlander is bound to an instance of `Credentials` and thus bound to a client (unless you share credentials between clients). If you need different credentials/regions/clients, you need multiple client instances with different credential instances.

src/aiodynamo/sign.py
ojii: pyperf tells me `Benchmark hidden because not significant (1): sign`

src/aiodynamo/credentials.py
ojii: actually, after sleeping on it, I realized that the decorator is per-class, not per-instance. I'll need to change this.

dimaqq:
🙇‍♂️

tests/unit/test_expressions.py
dimaqq: Eh to me (someone who didn't write the code) this looks roughly like this:
```
@pytest.mark.parametrize(
    "pe,expression,names",
    [
        %%%%%%%%%% ####### ###%###%###%##%###%
        %%%%%%%%%%%% ######## ####%####%####%#####%
            %%%%%%%%%%%%%
            ###########
            #####%####%#####%####%####%
```

I do blame amazon for the crazy syntax with all the oddball non-alpha characters...
TBH I don't know if how this could be improved.

One thing that was hard was to figure out which of the pytest parameters (pr, exp, nam) were inputs and which outputs.

tests/integration/test_client.py
```diff
@@ -204,22 +160,27 @@ def _get_tables(core):
         index += 1
 
 
-async def test_exists(client: Client):
-    name = str(uuid.uuid4())
+async def test_exists(client: Client, table_name_prefix: str):
+    name = table_name_prefix + str(uuid.uuid4())
```
dimaqq: maybe a bit OT... what do you think about `secrets.token_hex(4)` instead of UUID?

src/aiodynamo/utils.py
```diff
@@ -132,3 +54,80 @@ def deserialize(value: Dict[str, Any], numeric_type: Callable[[str], Any]) -> An
...
+                "BS",
+                [base64.b64encode(item).decode("ascii") for item in value if item],
+            )
+        else:
+            raise TypeError(
+                f"Sets which are not entirely numeric, strings or bytes are not supported. value: {value!r}"
+            )
+    elif isinstance(value, collections_abc.Mapping):
+        return "M", serialize_dict(value)
+    elif isinstance(value, list):
```
dimaqq: I wonder if some `collections_abc.Xxx` would be better than literal `list`, to cover, e.g. tuples and namedtuples and what not.

src/aiodynamo/utils.py
```diff
@@ -132,3 +54,80 @@ def deserialize(value: Dict[str, Any], numeric_type: Callable[[str], Any]) -> An
...
+        return "L", [item for item in map(serialize, value) if item is not None]
```
dimaqq: perhaps a small speed-up can be gained by:
```py
return "L", list(filter(partial(operator.is_not, None), map(serialize, value)))
```

dimaqq: Err, I've just tested it, looks like comprehension is actually a bit faster 😓

src/aiodynamo/http/base.py
```diff
@@ -0,0 +1,41 @@
...
+    @abc.abstractmethod
+    async def post(
+        self, *, url: URL, body: bytes, headers: Optional[Headers] = None
+    ) -> Dict[str, Any]:
+        """
+        Make a POST request and return the parsed JSON object.
+
+        Raise a RequestFailed exception if the request was not successful.
+        """
+        pass
```
dimaqq: I'm curious why the API has `get() -> bytes` but `post() -> json` ?

src/aiodynamo/expressions.py
dimaqq: Hmmm maybe they changed something.
In the past, dynamo was very pedantic.. it would only accept "optimally parenthesised" queries, meaning, for example, no redundant parentheses were allowed.

docs/changelog.rst
```diff
@@ -1,14 +1,20 @@
...
-20.2
+20.3
 ----
 
 Release Date: Unreleased
 
 * Added TTL support
 * Fixed a typo in ``delete_item``
 * Improved item deserialization performance
+* Improved overall client performance, especially for query, scan and count.
```
dimaqq: 💯I figure this could be emphasised more 💯

docs/concepts.rst
```diff
@@ -1,8 +1,18 @@
 Concepts
 ========
 
-Main differences to aiobotocore
--------------------------------
+Aiodynamo does not rely on boto3, botocore or aiobotocore. This is primarily done for performance reasons,
+as profiling showed that especially for querying data, boto libraries are very slow.
```
dimaqq: Grammar nazi: `... showed ... were ... slow`

docs/quickstart.rst
```diff
@@ -1,17 +1,11 @@
...
-
-aiodynamo requires Python 3.7 or higher.
-
-``pip install aiodynamo``
-
 
 Example
 -------
 
+``pip install aiodynamo[aiohttp]``
```
dimaqq: Do we recomment aiohttp or httpx?
I guess aiohttp is more... stable? but httpx is a bit faster? 🤔

docs/usage.rst
```diff
@@ -0,0 +1,88 @@
...
+
+The :py:class:`aiodynamo.client.Client` class takes three required and three optional arguments:
+
+1. An HTTP client adaptor, conforming to the :py:class:`aiodynamo.http.base.HTTP` interface.
+2. An instance of :py:class:`aiodynamo.credentials.Credentials` to authenticate with DynamoDB. You may use
+   ``Credentials.auto()`` to use the default loading strategy.
+3. The region your DynamoDB is in.
+4. An optional endpoint URL of your DynamoDB, as a :py:class:`yarl.URL` instance. Useful when using a local DynamoDB implementation such as dynalite or dynamodb-local.
+5. Which numeric type to use. This should be a callable which accepts a string as input and returns your numeric type as output. Defaults to ``float``.
+6. The throttling configuration to use. An instance of :py:class:`aiodynamo.models.ThrottleConfig`. By default, if the DynamoDB rate limit is exceeded, aiodynamo will attempt up to 5 times with increasing delays betwene requests.
```
dimaqq: spelling: between

src/aiodynamo/http/base.py
ojii: Yea that is indeed odd, but I felt for a specialized HTTP client interface it should be fine. The reason is that `post` is used for DynamoDB APIs and the return value is always JSON, so clients that have some form of built-in, possibly optimized, JSON parser can use that (maybe less copying of the response or something). The `get` API is used by credential fetchers and their responses are not always JSON, so that returns `bytes` instead. In the end, using this strange API is a bit nicer, so I think it's worth it.

src/aiodynamo/utils.py
ojii: 🤔 `Sequence` might indeed be better. (`Iterable`/`Iterator` might be a bit too wild). I used `list` here because that's what boto3 uses...

src/aiodynamo/client.py
dimaqq: My concern is that the top-level query API may be ineffecient.

For example, imagine that items are tiny like `{"R": 42}` and thus each page is like 100,000 items.

In that case client performing `async for item in query(): pass` may hit a bottleneck of asyncio async iteration rather than parsing dynamo response.

tests/integration/test_client.py
ojii: either is probably totally fine.

docs/quickstart.rst
ojii: I don't think we need to make a recommendation.

src/aiodynamo/client.py
dimaqq: For example:

```py
In [1]: async def foo():
   ...:     for i in range(10**6):
   ...:         yield i
   ...:

In [3]: async def asum():
   ...:     sum = 0
   ...:     async for i in foo():
   ...:         sum += i
   ...:     return sum
   ...:
```

This, when timed, gives me 0.17s per `await asum()`, which is 0.17 usec per `i`. Perhaps that's fast enough?

dimaqq: Certainly fast enough, compared to our... 10s for 35k rows... closing.

src/aiodynamo/credentials.py
```diff
@@ -0,0 +1,286 @@
...
+    max_attempts: int = 1
+    base_url: URL = URL("http://169.254.169.254")
+    disabled: bool = field(
+        default_factory=lambda: os.environ.get(
+            "AWS_EC2_METADATA_DISABLED", "false"
+        ).lower()
+        == "true"
+    )
+
+    def is_disabled(self) -> bool:
```
dimaqq: I wonder if there's any hint to detect that code is running on an instance.

Why it matters? If someone didn't configure anything (env, cont env), then default chain metadata thing falls back to instance metadata, and it will try to access 169.254.169.254, which is a reseved link-local address...

I dunno if someone could somehow hack the client.. 
Definitely, trying to get something from this address takes a while (socket timeout or http timeout).

```
> curl -v http://169.254.169.254
*   Trying 169.254.169.254...
* TCP_NODELAY set
* Connected to 169.254.169.254 (169.254.169.254) port 80 (#0)
> GET / HTTP/1.1
> Host: 169.254.169.254
> User-Agent: curl/7.64.1
> Accept: */*
>
* Empty reply from server
* Connection #0 to host 169.254.169.254 left intact
curl: (52) Empty reply from server
* Closing connection 0
```

docs/usage.rst
dimaqq:
```suggestion
6. The throttling configuration to use. An instance of :py:class:`aiodynamo.models.ThrottleConfig`. By default, if the DynamoDB rate limit is exceeded, aiodynamo will attempt up to 5 times with increasing delays between requests.
```

src/aiodynamo/credentials.py
ojii: if you worry (and know your env), use a concrete `Credentials` (such as `ContainerMetadataCredentials`) subclass rather than `Credentials.auto()`. if you're not on AWS, you should have the env vars set so this code never even runs.

src/aiodynamo/client.py
```diff
@@ -315,61 +399,68 @@ def table(self, name: TableName) -> Table:
...
+            wait_for_disabled = WaitConfig.default()
+        attempts = 0
+        while attempts < wait_for_disabled.max_attempts:
+            try:
+                await self.describe_table(table)
+            except TableNotFound:
+                return
+            attempts += 1
+            await asyncio.sleep(wait_for_disabled.retry_delay)
+        raise TableDidNotBecomeDisabled()
```
michaelaw320: some sort of generic async generator might work

pardon my pseudopython
```
some_generator(the_coro, max_attempts, delay):
attempts = 0
while attempts < max_attempts:
    yield the_coro()
    attempts += 1
    await asyncio.sleep(delay)
raise OutOfAttempts
```

in this case
```python3
try:
    async for state in some_generator(self.describe_table(table), wait_for_disabled.max_attempts, wait_for_disabled.retry_delay):
        pass
except TableNotFound:
    return
except OutOfAttempts:
    raise TableDidNotBecomeDisabled()
#Any other exception will be raised I assume?    
```

in other case the result can be compared with state and just stops the generator when desired state has been reached?

tests/integration/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@pytest.fixture
+def table_name_prefix() -> str:
+    return os.environ.get("DYNAMODB_TABLE_PREFIX", "")
+
+
+@pytest.fixture
+def endpoint():
+    if os.environ.get("TEST_ON_AWS", "false") == "true":
```
michaelaw320: how about 
```python3
if os.environ.get("TEST_ON_AWS"):
```

ojii: I think it's rare enough that being super explicit about what you're doing is fine. this is the tests after all.

src/aiodynamo/client.py
ojii: did something generator based, but a little different. what do you think of d60102c ?

Issue #21, opened 2020-03-18
fast: test (setup-ish) fails against AWS


dimaqq:
```
    async def test_query_with_limit(client: Client, fast_table: TableName):
        big = "x" * 20_000

>       await asyncio.gather(
            *(
                client.put_item(fast_table, {"h": "h", "r": str(i), "big": big})
                for i in range(100)
            )
        )

tests/integration/test_client.py:239:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/aiodynamo/client.py:483: in put_item
    resp = await self.send_request(action="PutItem", payload=payload)
src/aiodynamo/client.py:650: in send_request
    return await self.http.post(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = HTTPX(client=<httpx.client.AsyncClient object at 0x7fd9914d37f0>)

    async def post(
        self, *, url: URL, body: bytes, headers: Optional[Headers] = None
    ) -> Dict[str, Any]:
        response = await self.client.post(str(url), data=body, headers=headers)
        if response.status_code >= 400:
>           raise exception_from_response(response.status_code, await response.aread())
E           aiodynamo.errors.ProvisionedThroughputExceeded: {'__type': 'com.amazonaws.dynamodb.v20120810#ProvisionedThroughputExceededException', 'message': 'The level of configured provisioned throughput for the table was exceeded. Consider increasing your provisioning level with the UpdateTable API.'}
```

I've hacked it to create tables with 1000/1000 rcu (?) and yet, test cannot insert 100 items concurrently.
😠

ojii:
> I've hacked it to create tables with 1000/1000 rcu (?) and yet, test cannot insert 100 items concurrently.

is this on dynalite or real dynamodb?

ojii:
`test_query_with_limit` requires ~2100 Write Capacity Units (WCU = (size(item) / 1KB) * num_items), because the items are intentionally "big" (~20kb)

dimaqq:
real dynamo ofc, dynalite doesn't enforce limits :)

[...]

PR #22, opened 2020-03-18
try to test on aws with higher throughput

dimaqq:
Trying to address https://github.com/HENNGE/aiodynamo/issues/21

Still fails from my laptop...

### DIFF
--- a/src/aiodynamo/models.py
+++ b/src/aiodynamo/models.py
@@ -7,7 +7,8 @@ import random
 import time
 from dataclasses import dataclass
 from enum import Enum, unique
-from typing import AsyncIterable, Dict, List, Optional, Union
+from itertools import count
+from typing import AsyncIterable, Dict, Iterable, List, Optional, Union
 
 from .errors import Throttled
 
@@ -177,19 +178,29 @@ class WaitConfig:
             await asyncio.sleep(self.retry_delay)
 
 
+@dataclass(frozen=True)
 class ThrottleConfig(metaclass=abc.ABCMeta):
+    time_limit_secs: int = 60
+
     @classmethod
     def default(cls):
-        return DecorrelatedJitterThrottling()
+        return ExponentialBackoffThrottling()
 
     @abc.abstractmethod
-    async def attempts(self) -> AsyncIterable[None]:
+    def delays(self) -> Iterable[float]:
         raise NotImplementedError()
+
+    async def attempts(self) -> AsyncIterable[None]:
+        deadline = time.monotonic() + self.time_limit_secs
+        for delay in self.delays():
+            yield
+            if time.monotonic() > deadline:
+                raise Throttled()
+            await asyncio.sleep(delay)
 
 
 @dataclass(frozen=True)
 class DecorrelatedJitterThrottling(ThrottleConfig):
-    time_limit_secs: int = 60
     max_time_secs: int = 60
     base_delay_secs: int = 0.05
     max_delay_secs: int = 1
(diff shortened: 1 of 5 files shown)


src/aiodynamo/models.py
```diff
@@ -174,11 +177,36 @@ def default(cls):
...
+    async def attempts(self) -> AsyncIterable[None]:
+        raise NotImplementedError()
+
+
+@dataclass(frozen=True)
+class DecorrelatedJitterThrottling(ThrottleConfig):
+    time_limit_secs: int = 60
+    max_time_secs: int = 60
+    base_delay_secs: int = 0.05
+    max_delay_secs: int = 1
```
dimaqq: I think we need to pass something to Deco, so that it knows how much we're trying to do:
* inconsistent read costs ½RCU
* [consistent read]
* write costs 4RCU or something
* query costs anywhere between 125RCU (1MB/4K/2) to 100KRCU (tiny items)?
--
* table could be provisioned however...
* there could be N competing users (e.g. containers)
--
I don't have a good idea how to predict how long to wait for dynamo; but I figure that get and scan should have different initial conditions...

dimaqq: Then again, it's nicely exponential, with ╳1.5 per iteration, so it only takes ~8 steps to get from .05s to 1s, should be OK?

tests/integration/conftest.py
```diff
@@ -14,38 +17,59 @@ def table_name_prefix() -> str:
...
+
+@pytest.fixture
+async def table(
+    client: Client, table_factory: Callable[[Optional[Throughput]], Awaitable[str]]
+) -> str:
+    name = await table_factory()
+    try:
+        yield name
+    finally:
+        await client.delete_table(name)
```
dimaqq: I wonder if it's a good cause for walrus:
```py
try:
    yield (name := await table_factory())
finally:
   await delete(name)
```
---
Though, yeah, technically, it's not quite correct, in case table creation fails 🤔

src/aiodynamo/models.py
```diff
@@ -203,10 +214,14 @@ def delays(self):
...
+@dataclass(frozen=True)
+class ExponentialBackoffThrottling(ThrottleConfig):
+    base_delay_secs: int = 2
+    max_delay_secs: int = 20
+
+    def delays(self) -> Iterable[float]:
+        for attempt in count():
+            yield min(
+                random.random() * (self.base_delay_secs ** attempt), self.max_delay_secs
+            )
```
dimaqq: I consider `Deco` already exponential...

tests/integration/conftest.py
ojii: I think the walrus should be used sparingly because it's quite weird. I don't find your version easier to read.

src/aiodynamo/models.py
ojii: what is `Deco`?

dimaqq: your decompressed random throttler thing... DecorrelatedBlahBlahThrottle.

src/aiodynamo/models.py
dimaqq: here too, I meant `DecorrelatedBackoffThrottling` is already exponential... I think 💭 🤔

ojii: they behave different though. the Decorelated... I got from their blog post about throttling. this one is from botocore. naming isn't great, but let me run the tests with them to see how they work first.

ojii: `ExponentialBackoffThrottling` passes the tests suite (with default args)

ojii: `DecorelatedJitterThrottling` also passes.

ojii: and just to make sure, disabling throttling makes the tests fail: ```python
class NoThrottle(ThrottleConfig):
    def delays(self):
        yield 0
        raise Exception('no throttling!')
```

PR #23, opened 2020-03-23
more logging, better token expiration handling

ojii:
seems like boto is using 10 minutes for mandatory token refresh, so I adopted the same deadline. 

Added a first class Exception for expired tokens
Retry on expired tokens if the token could be invalidated
Added more logging


ojii:
> Maybe don't log the value of the `key`?
> (I'm not clear which part is secret there)

the secret parts are stripped out, `repr(key)` only shows the access key id

ojii:
> I think, later on, you may want to reduce `logging.info` only to things that have happened.

I don't actually know why i used info in one file and debug in the other. made them all debug now.

PR #36, opened 2020-04-08
wrap client errors

ojii:
wrap client errors for retries to work.

### DIFF
--- a/src/aiodynamo/http/aiohttp.py
+++ b/src/aiodynamo/http/aiohttp.py
@@ -1,18 +1,39 @@
 from dataclasses import dataclass
-from typing import Any, Dict, Mapping, Optional
+from functools import wraps
+from typing import Any, Dict, Optional
 
 from aiodynamo.types import Timeout
-from aiohttp import ClientSession
+from aiohttp import ClientError, ClientResponseError, ClientSession
 from yarl import URL
 
 from ..errors import exception_from_response
 from .base import HTTP, Headers, RequestFailed
 
 
+def wrap_errors(coro):
+    @wraps(coro)
+    async def wrapper(*args, **kwargs):
+        try:
+            return await coro(*args, **kwargs)
+        except ClientResponseError as exc:
+            raise RequestFailed(
+                exc.request_info.url,
+                exc.status,
+                exc.message,
+                exc.request_info.headers,
+                None,
+            )
+        except ClientError:
+            raise RequestFailed()
+
+    return wrapper
+
+
 @dataclass(frozen=True)
 class AIOHTTP(HTTP):
     session: ClientSession
 
+    @wrap_errors
     async def get(
         self, *, url: URL, headers: Optional[Headers] = None, timeout: Timeout
     ) -> bytes:
@@ -25,6 +46,7 @@ class AIOHTTP(HTTP):
                 )
             return await response.read()
 
+    @wrap_errors
     async def post(
         self, *, url: URL, body: bytes, headers: Optional[Headers] = None
     ) -> Dict[str, Any]:
--- a/src/aiodynamo/http/aiohttp.py
+++ b/src/aiodynamo/http/aiohttp.py
@@ -3,7 +3,7 @@ from functools import wraps
 from typing import Any, Dict, Optional
 
 from aiodynamo.types import Timeout
-from aiohttp import ClientError, ClientResponseError, ClientSession
+from aiohttp import ClientError, ClientSession
 from yarl import URL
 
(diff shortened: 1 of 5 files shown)


src/aiodynamo/http/aiohttp.py
```diff
@@ -1,18 +1,39 @@
...
+        except ClientResponseError as exc:
+            raise RequestFailed(
+                exc.request_info.url,
+                exc.status,
+                exc.message,
+                exc.request_info.headers,
+                None,
+            )
+        except ClientError:
+            raise RequestFailed()
```
dimaqq, line 27: I wonder if there's some info...
Would something like `... as e: raise R(str(e))` or `repr` be useful?
For example so that user can distinguish error cause, dns vs connection.

On the other hand, exceptions are chained by default, so perhaps there's enough information there already? 🤔

src/aiodynamo/http/httpx.py
```diff
@@ -1,19 +1,38 @@
 import json
 from dataclasses import dataclass
+from functools import wraps
 from typing import Any, Dict, Optional
 
 from aiodynamo.types import Timeout
-from httpx import AsyncClient
+from httpx import AsyncClient, HTTPError
```
dimaqq, line 7: I wonder if `httpx` passes underlying socket errors through

src/aiodynamo/http/aiohttp.py
ojii: > On the other hand, exceptions are chained by default, so perhaps there's enough information there already? 🤔

that was my thinking.

src/aiodynamo/http/httpx.py
ojii: judging solely based on the names of the exceptions, it looks like it https://github.com/encode/httpx/blob/master/httpx/_exceptions.py

src/aiodynamo/http/aiohttp.py
```diff
@@ -1,18 +1,39 @@
 from dataclasses import dataclass
-from typing import Any, Dict, Mapping, Optional
+from functools import wraps
+from typing import Any, Dict, Optional
 
 from aiodynamo.types import Timeout
-from aiohttp import ClientSession
+from aiohttp import ClientError, ClientResponseError, ClientSession
```
dimaqq, line 6: I wonder if perhaps `aiohttp` passed some underlying socket errors though...
e.g. `gaierror: [Errno -3] Try again` from `socket.getaddrinfo`
please test with wifi disconnected or something...

src/aiodynamo/http/httpx.py
dimaqq: I guess we too may need something like this:
https://github.com/encode/httpx/blob/53804173bb734ce94d13d1b34c0ebaab46abd8c3/tests/dispatch/test_http2.py#L155

src/aiodynamo/http/aiohttp.py
ojii: [looks like they do](https://github.com/aio-libs/aiohttp/blob/5a0da90418ffc8cfcd8190bf8027203770a3162e/aiohttp/client_exceptions.py#L101-L109)

src/aiodynamo/http/httpx.py
```diff
@@ -1,19 +1,38 @@
...
+    async def wrapper(*args, **kwargs):
+        try:
+            return await coro(*args, **kwargs)
+        except HTTPError as exc:
+            raise RequestFailed(
+                exc.request and exc.request.url,
+                exc.response and exc.response.status_code,
+                exc.response and await exc.response.aread(),
+                exc.request and exc.request.headers,
+                exc.request and await exc.request.aread(),
```
dimaqq, line 25: Request is ours, so I guess it should be OK?
Do we want to the body for anything though, are we logging it? Is it safe to log, i.e. there are no keys there?

src/aiodynamo/http/httpx.py
```diff
@@ -1,19 +1,38 @@
...
+def wrap_errors(coro):
+    @wraps(coro)
+    async def wrapper(*args, **kwargs):
+        try:
+            return await coro(*args, **kwargs)
+        except HTTPError as exc:
+            raise RequestFailed(
+                exc.request and exc.request.url,
+                exc.response and exc.response.status_code,
+                exc.response and await exc.response.aread(),
```
dimaqq, line 23: I wonder if it's "safe" to read the response if there's an error.
I mean, are all the errors like 500 + body?
Or are some errors like `ConnectionResetError`?

PR #38, opened 2020-04-08
make RequestFailed simpler, add retry

ojii:
Remove all the extra info from `RequestFailed`, the caller can still get it from the chained exception. Add super simple retry logic to `send_request` retrying all `RequestFailed` errors.

### DIFF
--- a/src/aiodynamo/http/aiohttp.py
+++ b/src/aiodynamo/http/aiohttp.py
@@ -33,7 +33,7 @@ class AIOHTTP(HTTP):
             method="GET", url=url, headers=headers, timeout=timeout
         ) as response:
             if response.status >= 400:
-                raise RequestFailed(response.status)
+                raise RequestFailed()
             return await response.read()
 
     @wrap_errors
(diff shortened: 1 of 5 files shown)


src/aiodynamo/http/aiohttp.py
```diff
@@ -41,9 +33,7 @@ class AIOHTTP(HTTP):
             method="GET", url=url, headers=headers, timeout=timeout
         ) as response:
             if response.status >= 400:
-                raise RequestFailed(
-                    url, response.status, await response.read(), headers, None
-                )
+                raise RequestFailed(response.status)
```
dimaqq, line 36: How does dynamo server signal logical errors, is it not `400` status code?

dimaqq: Ahh, sorry I was looking at wrong method, I can see that handled below in `post`.

ojii: oops shouldn't include status, forgot to remove it here

dimaqq:
👏

PR #46, opened 2020-04-22
Improved the way the client handles no key being returned from credentials

ojii:
Currently, when no credentials can be found, the client fails with a strange `AttributeError`. This change will make the client retry on credentials resolving failure and raise an appropriate error if no credentials could ever be fetched.

### DIFF
--- a/src/aiodynamo/errors.py
+++ b/src/aiodynamo/errors.py
@@ -5,6 +5,10 @@ class AIODynamoError(Exception):
     pass
 
 
+class NoCredentialsFound(Exception):
+    pass
+
+
 class EmptyItem(AIODynamoError):
     pass
 
--- a/src/aiodynamo/errors.py
+++ b/src/aiodynamo/errors.py
@@ -5,7 +5,7 @@ class AIODynamoError(Exception):
     pass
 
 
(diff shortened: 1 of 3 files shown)


src/aiodynamo/errors.py
```diff
@@ -5,6 +5,10 @@ class AIODynamoError(Exception):
     pass
 
 
+class NoCredentialsFound(Exception):
```
dimaqq, line 8: I'm thinking about pros/cons of inheriting from `Exception` and not `AIODynamoError`...

I wonder what's least unexpected:
* first failed request almost killing user program
* or each request failing with same error

My weak preference would be for the new exception to inherit from `AIODynamoError` because:
1. user program may be using several libraries together, and `aiodynamo` should not be "special".
2. if a user program only uses `aiodynamo` but wants to exit quickly if there are no credentials, it can still catch this type

ojii: > I'm thinking about pros/cons of inheriting from `Exception` and not `AIODynamoError`...

That was a complete brainfart from me, I intended to inherit from the root exception but for some reason wrote `Exception`.

PR #50, opened 2020-05-20
remove empty string handling

### LINKED ISSUE #48: Support empty strings
Empty strings (and bytes) are [now allowed](https://aws.amazon.com/about-aws/whats-new/2020/05/amazon-dynamodb-now-supports-empty-values-for-non-key-string-and-binary-attributes-in-dynamodb-tables/) except in key/index fields. aiodynamo should support this.

Outstanding questions are:

* Should aiodynamo try to check key fields are not empty? Currently keys are not handled special (eg their types are not checked even though only S, B and N are allowed)
* How would aiodynamo know which fields are keys? In `update_item` that's trivial (though indices cannot be detected) but `put_item` the user would have to supply the keys somehow.
* Maybe this should be something only handled on `Table`, not `Client`, by having the user pass a list of key/index fields to the constructor.

Outstanding problems:

* dynalite does not support this yet

ojii:
Fixes #48, removes special handling for empty strings and bytes.

Sadly, DynamoDB emulators seem to not yet support this so tests are marked as xfail for now.

### DIFF
--- a/tests/integration/test_client.py
+++ b/tests/integration/test_client.py
@@ -10,6 +10,7 @@ from aiodynamo.errors import (
     NoCredentialsFound,
     TableNotFound,
     UnknownOperation,
+    ValidationException,
 )
 from aiodynamo.expressions import F, HashKey, RangeKey
 from aiodynamo.http.base import HTTP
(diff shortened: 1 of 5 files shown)


tests/integration/test_client.py
```diff
@@ -217,7 +220,7 @@ async def test_empty_string(client: Client, table: TableName):
 
 
 async def test_empty_item(client: Client, table: TableName):
-    with pytest.raises(EmptyItem):
+    with pytest.raises(ValidationException):
```
dimaqq, line 223: Does it know that `h` and `r` are special? 🤔

dimaqq:
Document that empty strings are now supported:
* changelog
* exception that's no longer raised, if we had such

tests/integration/test_client.py
ojii: thee client doesn't. the server does. `ValidationException` comes from the server.

PR #56, opened 2020-10-01
Better client errors

### LINKED ISSUE #55: Improve HTTP client error reporting
When HTTP client wrappers fail for any reason, they raise a `aiodynamo.http.base.RequestFailed` error to signal to the rest  of aiodynamo that this was not a DynamoDB side issue and that a request should likely be retried. However, right now it doesn't do a good job at reporting *why* the HTTP Client failed, making diagnosing these errors hard/impossible.

ojii:
Explicitly chains client errors (`RequestFailed`) and gives them an (optional) string reason.

Since I needed `StaticCredentials` for easy testing, I've added that too, as it can be handy. And since I touched Credentials, I've also improved their documentation. Sorry for the mixing of concerns.

Fixes #55

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -7,7 +7,9 @@ Custom HTTP client adaptor
 If for some reason you want to use neither aiohttp nor httpx, you can adapt your
 own HTTP client for usage with aiodynamo. To do so, create a class that conforms
 to the :py:class:`aiodynamo.http.base.HTTP` interface. Errors should be wrapped in
-a :py:exc:`aiodynamo.http.base.RequestFailed`.
+a :py:exc:`aiodynamo.http.base.RequestFailed`. When raising a :py:exc:`aiodynamo.http.base.RequestFailed`
+due to an exception in the HTTP client, you **should explicitly chain the exceptions**
+using ``raise RequestFailed(reason) from client_exception``.
 
 .. autoclass:: aiodynamo.http.base.HTTP
     :members:
--- a/tests/unit/test_http_clients.py
+++ b/tests/unit/test_http_clients.py
@@ -1,0 +1,40 @@
+import asyncio
+from contextlib import asynccontextmanager
+
+import pytest
+from aiodynamo.client import Client
+from aiodynamo.credentials import Key, StaticCredentials
+from aiodynamo.http.aiohttp import AIOHTTP
+from aiodynamo.http.base import RequestFailed
+from aiodynamo.models import ExponentialBackoffThrottling
+from yarl import URL
+
+pytestmark = [pytest.mark.asyncio]
+
+
+async def test_aiohttp_request_failed():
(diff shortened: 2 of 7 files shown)


tests/unit/test_http_clients.py
```diff
@@ -0,0 +1,40 @@
...
+        @asynccontextmanager
+        async def request(self, *args, **kwargs):
+            raise exc
+            yield
+
+    http = AIOHTTP(FakeSession())
+
+    with pytest.raises(RequestFailed) as error:
+        await http.get(url=URL("http://foo.invalid.com"), timeout=100)
+    assert error.value.__cause__ is exc
```
dimaqq, line 28: I wonder if it's enough to support both `__cause__` and `__context__` as in PEP-3134:
https://www.python.org/dev/peps/pep-3134/#enhanced-reporting

docs/advanced.rst
```diff
@@ -7,7 +7,9 @@ Custom HTTP client adaptor
 If for some reason you want to use neither aiohttp nor httpx, you can adapt your
 own HTTP client for usage with aiodynamo. To do so, create a class that conforms
 to the :py:class:`aiodynamo.http.base.HTTP` interface. Errors should be wrapped in
-a :py:exc:`aiodynamo.http.base.RequestFailed`.
+a :py:exc:`aiodynamo.http.base.RequestFailed`. When raising a :py:exc:`aiodynamo.http.base.RequestFailed`
+due to an exception in the HTTP client, you **should explicitly chain the exceptions**
+using ``raise RequestFailed(reason) from client_exception``.
```
dimaqq, line 12: I'd rephrase the start of the paragraph, e.g.:

Aiodynamo comes with built-in support for `aiohttp` and `httpx` transports, and you can create your own by implementing the XX interface. 

Re explicitly chaining exceptions: please explain.
If it's only to help end-user debug errors, then I'd rather **recommend** to preserve context (or cause)?

tests/unit/test_http_clients.py
ojii: From how I  understand it, `__cause__` is for explicit chaining,  `__context__` for implicit. Since this PR changes to explicit chaining, checking for `__cause__` seems reasonable.

dimaqq: I guess what puzzles me is why does `aiodynamo` care?
The library evaluates neither `__cause__` nor `__context__`, so what's the rationale for the requirement?
In other words, why can't the transport use implicit (default) chaining?

ojii:
I think after all this doesn't make things much better, so I'm closing it for now.

dimaqq:
Does it make sense to provide `StaticCredentials` somehow? Or is that just silly?

ojii:
> Does it make sense to provide `StaticCredentials` somehow? Or is that just silly?

It's kinda useful for writing tests.

PR #60, opened 2020-10-13
Add full typing

ojii:
Passes mypy with strict enabled, but needs some ugly workarounds and way too much `Any` (mostly in internal code) for my taste. But it should still be useful for users since this would make their mypy check their usage of aiodynamo.

### DIFF
--- a/src/aiodynamo/client.py
+++ b/src/aiodynamo/client.py
@@ -51,14 +51,14 @@ class TimeToLive:
 
     async def enable(
         self, attribute: str, *, wait_for_enabled: Union[bool, WaitConfig] = False
-    ):
+    ) -> None:
         await self.table.client.enable_time_to_live(
             self.table.name, attribute, wait_for_enabled=wait_for_enabled
         )
 
     async def disable(
         self, attribute: str, *, wait_for_disabled: Union[bool, WaitConfig] = False
-    ):
+    ) -> None:
         await self.table.client.disable_time_to_live(
             self.table.name, attribute, wait_for_disabled=wait_for_disabled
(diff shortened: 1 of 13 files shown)


src/aiodynamo/credentials.py
```diff
@@ -198,7 +203,9 @@ async def get_key(self, http: HTTP) -> Optional[Key]:
                 self._refresher = asyncio.create_task(self._refresh(http))
             else:
                 logger.debug("already refreshing")
-        return self._metadata and self._metadata.key
+        if self._metadata:
+            return self._metadata.key
+        return None
```
dimaqq, line 208: Does `mypy` force you to do this? 😱

src/aiodynamo/client.py
```diff
@@ -669,6 +685,7 @@ async def send_request(
             if failed is not None:
                 raise failed
             raise
+        raise BrokenThrottleConfig()
```
dimaqq, line 688: Is this to somehow handle empty generator throttle, or to silence mypy?

dimaqq (approved):

src/aiodynamo/client.py
ojii: primarily: silence mypy. However, I realized that if a custom throttle config for some reason immediately exits, it would actually also reach this point and indicate broken code, so raising an exception seemed reasonable after all here.

src/aiodynamo/credentials.py
ojii: it didn't understand the `and` shortcut 😭

PR #62, opened 2020-10-14
Role ARN support in InstanceMetadata Credentials

### LINKED ISSUE #61: Url gets invalid if instance metadata return role arn instead of just a role name
Hey

I found another problem while using this library.

When using `InstanceMetadataCredentials` there is an error thrown from httpx:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/aiodynamo/http/httpx.py", line 18, in wrapper
    return await coro(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/aiodynamo/http/httpx.py", line 33, in get
    response = await self.client.get(str(url), headers=headers, timeout=timeout)
  File "/usr/local/lib/python3.8/site-packages/httpx/client.py", line 1236, in get
    return await self.request(
  File "/usr/local/lib/python3.8/site-packages/httpx/client.py", line 1086, in request
    request = self.build_request(
  File "/usr/local/lib/python3.8/site-packages/httpx/client.py", line 200, in build_request
    url = self.merge_url(url)
  File "/usr/local/lib/python3.8/site-packages/httpx/client.py", line 220, in merge_url
    url = self.base_url.join(relative_url=url)
  File "/usr/local/lib/python3.8/site-packages/httpx/models.py", line 229, in join
    return URL(relative_url)
  File "/usr/local/lib/python3.8/site-packages/httpx/models.py", line 108, in __init__
    raise InvalidURL("No host included in URL.")
httpx.exceptions.InvalidURL: No host included in URL.
```

I could track the problem down to [this line](https://github.com/HENNGE/aiodynamo/blob/edc7b3379257a6f1d6c2aca29e5617e760c9c037/src/aiodynamo/credentials.py#L310)

```python
>>> role_url = URL("http://169.254.169.254/latest/meta-data/iam/security-credentials/")
>>> role = "test-role"
>>> role_url.join(URL(role))
URL('http://169.254.169.254/latest/meta-data/iam/security-credentials/test-role')
>>> role = "arn:aws:iam::1234567890:role/test-role"
>>> role_url.join(URL(role))
URL('arn:aws:iam::1234567890:role/test-role') # expected would be URL('http://169.254.169.254/latest/meta-data/iam/security-credentials/arn:aws:iam::1234567890:role/test-role')
```

My use case is that the request gets intercepted with [KIAM](https://github.com/uswitch/kiam) and my code is supposed to assume this role. I tried the urls with curl and they worked as expected.

ojii:
Fixes #61

### DIFF
--- a/src/aiodynamo/credentials.py
+++ b/src/aiodynamo/credentials.py
@@ -159,7 +159,7 @@ class MetadataCredentials(Credentials, metaclass=abc.ABCMeta):
         max_attempts: int,
         url: URL,
         timeout: Timeout,
-        headers: Optional[Headers] = None
+        headers: Optional[Headers] = None,
     ) -> bytes:
         for attempt in range(max_attempts):
             try:
@@ -307,7 +307,9 @@ class InstanceMetadataCredentials(MetadataCredentials):
                 timeout=self.timeout,
             )
         ).decode("utf-8")
-        credentials_url = role_url.join(URL(role))
+        credentials_url = self.base_url.with_path(
+            f"/latest/meta-data/iam/security-credentials/{role}"
+        )
         raw_credentials = await self.fetch_with_retry(
             http=http,
             max_attempts=self.max_attempts,
--- a/tests/unit/test_credentials.py
+++ b/tests/unit/test_credentials.py
@@ -62,7 +62,7 @@ async def instance_metadata_server():
     app.add_routes(
         [
             web.get(
-                "/latest/meta-data/iam/security-credentials/{role}",
+                "/latest/meta-data/iam/security-credentials/{role:.*}",
                 ims.credentials_handler,
             )
         ]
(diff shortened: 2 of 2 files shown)


src/aiodynamo/credentials.py
```diff
@@ -307,7 +307,9 @@ async def fetch_metadata(self, http: HTTP) -> Metadata:
                 timeout=self.timeout,
             )
         ).decode("utf-8")
-        credentials_url = role_url.join(URL(role))
+        credentials_url = self.base_url.with_path(
+            f"/latest/meta-data/iam/security-credentials/{role}"
```
dimaqq, line 311: maybe unrelated:
I went to check what `role` might be... just in case it's smth like `../../../path-to-exfiltrate-data`
And I realised that `(aiohttp|httpx).get()` protects against status codes >=400, but not 302 or 201 or 204.
So, hypothetically, `role` could be empty or even random text.
I dunno what could break in that case...

tests/unit/test_credentials.py
```diff
@@ -96,14 +96,15 @@ async def test_env_credentials(monkeypatch, http):
     assert key.token == "token"
 
 
-async def test_ec2_instance_metdata_credentials(http, instance_metadata_server):
+@pytest.mark.parametrize("role", ["role", "arn:aws:iam::1234567890:role/test-role"])
```
dimaqq, line 99: Does the new test only check that "client doesn't crash"?

dimaqq (approved):
A release note would be good.

tests/unit/test_credentials.py
ojii: yes

src/aiodynamo/credentials.py
ojii: 🤔 if you compromised the instance metadata endpoint somehow, you can already hijack the first request (`"/latest/meta-data/iam/security-credentials/"`) so I don't think this changes much.

Issue #65, opened 2020-11-25
What is "Candidate" ?


misogihagi:
when I tried 

```python3
from aiodynamo.client import Client
from aiodynamo.credentials import Credentials
from aiodynamo.http.aiohttp import AIOHTTP
from aiohttp import ClientSession
import asyncio

async def main():
    async with ClientSession() as session:
        client = Client(AIOHTTP(session), Credentials.auto(), "ap-northeast-1")
        table = client.table("table")
        await table.put_item(
          {
                'id':0,
                'data':"000"
            }
        )
        
asyncio.run(main())

```

and terminal showed below

> Traceback (most recent call last):
>   File "C:\Users\user\AppData\Local\Programs\Python\Python39\lib\site-packages\aiodynamo\credentials.py", line 101, in get_key
>     key = await candidate.get_key(http)
>   File "C:\Users\user\AppData\Local\Programs\Python\Python39\lib\site-packages\aiodynamo\credentials.py", line 215, in get_key
>     await self._refresher
>   File "C:\Users\user\AppData\Local\Programs\Python\Python39\lib\site-packages\aiodynamo\credentials.py", line 239, in _refresh
>     self._metadata = await self.fetch_metadata(http)
>   File "C:\Users\user\AppData\Local\Programs\Python\Python39\lib\site-packages\aiodynamo\credentials.py", line 338, in fetch_metadata
>     await self.fetch_with_retry(
>   File "C:\Users\user\AppData\Local\Programs\Python\Python39\lib\site-packages\aiodynamo\credentials.py", line 196, in fetch_with_retry
>     raise TooManyRetries()
> aiodynamo.credentials.TooManyRetries
> Candidate InstanceMetadataCredentials(timeout=1, max_attempts=1, base_url=URL('http://169.254.169.254'), disabled=False) failed

I think this problem because of my networking

> ping 169.254.169.254
> 
> Pinging 169.254.169.254 with 32 bytes of data:
> PING: transmit failed.
> PING: transmit failed.
> PING: transmit failed.
> PING: transmit failed.
> 
> Ping statistics for 169.254.169.254:
>     Packets: Sent = 4, Received = 0, Lost = 4 (100% loss),
> 

But according to boto3 available,  I think it doesn't need this process.

dimaqq:
`aiodynamo` tries several ways (candidates) to get credentials to the server.
That particular ip address is a well-know value accessible either from docker container or aws instance (sorry I dunno which).
[...]

PR #66, opened 2020-11-28
credential file support

misogihagi:
#65 
brief implement without any profiles


ojii:
@misogihagi thank you for the pull request.

Reading up on how AWS SDKs load credentials from files, it looks like it's quite a bit more complicated, so I will make modifications based off your branch to support it more completely. Thank you for your patience while I work on that.

PR #69, opened 2021-01-08
Add table class to docs

hanscg:
Added table class to docs since this were the part where I was the most confused about. 
I'm thinking of adding sample code for query later.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -59,6 +59,46 @@ In case you want to explicitly pass the credentials from Python, use :py:class:`
     :members:
     :undoc-members:
 
+The ``Table`` class
+-------------------
+
+After instantiating client, set the table to be used by calling the :py:meth:`aiodynamo.client.Client.table` method.
+The method takes the table name string as its only argument.
+
+.. automethod:: aiodynamo.client.Client.table
+
+Modify table
+~~~~~~~~~~~~
+
+These methods can be used to modify the table.
+
+.. automethod:: aiodynamo.client.Table.exists
+.. automethod:: aiodynamo.client.Table.create
+.. automethod:: aiodynamo.client.Table.delete
+
+Modify items
+~~~~~~~~~~~~
+
+These methods can be used to modify items in the table.
+
+.. automethod:: aiodynamo.client.Table.put_item
+.. automethod:: aiodynamo.client.Table.update_item
+.. automethod:: aiodynamo.client.Table.delete_item
+
+Read items
+~~~~~~~~~~
+
+These methods can be used to return items from the table.
+
+.. automethod:: aiodynamo.client.Table.get_item
+
+|
+
+.. automethod:: aiodynamo.client.Table.query
+
+|
+
+.. automethod:: aiodynamo.client.Table.scan
 
 The ``F`` class
 ---------------
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -62,43 +62,56 @@ In case you want to explicitly pass the credentials from Python, use :py:class:`
 The ``Table`` class
 -------------------
 
-After instantiating client, set the table to be used by calling the :py:meth:`aiodynamo.client.Client.table` method.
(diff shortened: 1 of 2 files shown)


docs/usage.rst
```diff
@@ -59,6 +59,46 @@ In case you want to explicitly pass the credentials from Python, use :py:class:`
     :members:
     :undoc-members:
 
+The ``Table`` class
+-------------------
+
+After instantiating client, set the table to be used by calling the :py:meth:`aiodynamo.client.Client.table` method.
```
dimaqq, line 65: Somehow... the language could be better...

src/aiodynamo/client.py
```diff
@@ -73,6 +73,9 @@ class Table:
     name: TableName
 
     async def exists(self) -> bool:
+        """
+        Checks that the table exist.
```
dimaqq, line 77:
```suggestion
        Checks if the table exists.
```

dimaqq: I think that Python docs use the imperative mood, for example, `os.path.exists(...)`:
>     exists(path)
>        Test whether a path exists.  Returns False for broken symbolic links

docs/usage.rst
```diff
@@ -59,6 +59,46 @@ In case you want to explicitly pass the credentials from Python, use :py:class:`
...
+.. automethod:: aiodynamo.client.Table.delete_item
+
+Read items
+~~~~~~~~~~
+
+These methods can be used to return items from the table.
+
+.. automethod:: aiodynamo.client.Table.get_item
+
+|
```
dimaqq, line 95: What does the bar do?

ojii: it inserts empty space in the output to make it less cluttered.


in sphinx: 

```
foo
|
bar
```

is similar to html:

```
<p>foo</p>
<p>bar</p>
```

whereas sphinx:

```
foo

bar
```

would be (irregardless of how many newlines between foo and bar) 

```
foo<br>bar
```

dimaqq:
poke

dimaqq:
@ojii I guess we have to pick this up...

ojii:
> @ojii I guess we have to pick this up...

I know, I just have some doubts about the usefulness of documentation that says that `delete_table` "Deletes the table". And I especially don't think those docstrings are useful, so moving the explanation (especially of the arguments to the methods) should be in the docs imho, not the code.

hanscg:
I'm sorry for leaving this for so long. 
I'll try to fix it according to your reviews next Monday.

docs/usage.rst
```diff
@@ -59,6 +59,46 @@ In case you want to explicitly pass the credentials from Python, use :py:class:`
     :members:
     :undoc-members:
 
+The ``Table`` class
+-------------------
+
```
ojii, line 64: please add a note that the `Table` class should not be instantiated directly, but always via Client.table.

docs/usage.rst
```diff
@@ -59,6 +59,46 @@ In case you want to explicitly pass the credentials from Python, use :py:class:`
...
+After instantiating client, set the table to be used by calling the :py:meth:`aiodynamo.client.Client.table` method.
+The method takes the table name string as its only argument.
+
+.. automethod:: aiodynamo.client.Client.table
+
+Modify table
+~~~~~~~~~~~~
+
+These methods can be used to modify the table.
```
ojii, line 73: this says the same as the "Modify table" header, so it's not needed

ojii (changes requested):
maybe instead of the docstrings (which more or less say what the method name says), linking to the DynamoDB API docs (in the docs, not the code) might be more useful? In most cases, things map 1-to-1, so only the differences should be explained in detail in the docs. Off the top of my head, those are the auto-depagination in query/scan and the wait_for... in create/delete.

Also, a lot of datastructures used in arguments are not documented, they should at least get an autodoc.

ojii:
thank you, merged.

PR #72, opened 2021-05-21
Add API to query/scan single DynamoDB pages

ojii:
Adds two new APIs to both `Client` and `Table`: `query_single_page` and `scan_single_page`, both taking the same arguments as `query` and `scan` respectively. Unlike `query` and `scan`, they do not return an async iterator over items, but instead a `Page` which has a `List` of items and the `LastEvaluatedKey`. They can be used in cases where manual de-pagination is desired, such as returning items from DynamoDB via a paginated API.

### DIFF
--- a/src/aiodynamo/client.py
+++ b/src/aiodynamo/client.py
@@ -29,6 +29,7 @@ from .models import (
     KeySpec,
     KeyType,
     LocalSecondaryIndex,
+    Page,
     ReturnValues,
     Select,
     StreamSpecification,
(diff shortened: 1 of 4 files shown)


hanscg (commented):
I don't see any problem as far as I look ...
Maybe need some test for scan single page?

dimaqq:
Add new methods to the docs too 🙏🏿

src/aiodynamo/client.py
```diff
@@ -515,39 +578,100 @@ async def query(
...
+        filter_expression: Optional[Condition] = None,
+        scan_forward: bool = True,
+        index: Optional[str] = None,
+        limit: Optional[int] = None,
+        projection: Optional[ProjectionExpression] = None,
+        select: Select = Select.all_attributes,
+    ) -> Page:
+        """
+        Query a single DynamoDB page.
+        To automatically handle pagination, uses query(...) instead.
```
dimaqq, line 617:
```suggestion
        To automatically handle pagination, use query(...) instead.
```

src/aiodynamo/client.py
```diff
@@ -556,28 +680,35 @@ async def scan(
         start_key: Optional[Dict[str, Any]] = None,
         projection: Optional[ProjectionExpression] = None,
         filter_expression: Optional[Condition] = None,
-    ) -> AsyncIterator[Item]:
+    ) -> Page:
+        """
+        Scan a single DynamoDB page.
+        To automatically handle pagination, uses scan(...) instead.
```
dimaqq, line 686:
```suggestion
        To automatically handle pagination, use scan(...) instead.
```

src/aiodynamo/client.py
```diff
@@ -160,6 +167,34 @@ def query(
...
+        filter_expression: Optional[Condition] = None,
+        scan_forward: bool = True,
+        index: Optional[str] = None,
+        limit: Optional[int] = None,
+        projection: Optional[ProjectionExpression] = None,
+        select: Select = Select.all_attributes,
+    ) -> Page:
+        """
+        Query a single DynamoDB page.
+        To automatically handle pagination, uses query(...) instead.
```
dimaqq, line 184:
```suggestion
        To automatically handle pagination, use query(...) instead.
```

src/aiodynamo/client.py
```diff
@@ -178,6 +219,28 @@ def scan(
...
+        *,
+        index: Optional[str] = None,
+        limit: Optional[int] = None,
+        start_key: Optional[Dict[str, Any]] = None,
+        projection: Optional[ProjectionExpression] = None,
+        filter_expression: Optional[Condition] = None,
+    ) -> Page:
+        """
+        Scan a single DynamoDB page.
+        To automatically handle pagination, uses scan(...) instead.
```
dimaqq, line 233:
```suggestion
        To automatically handle pagination, use scan(...) instead.
```

PR #76, opened 2021-07-02
fix byte encoding for benchmark dummy data util

### LINKED ISSUE #75: Deserialize benchmarks broken
Benchmarks broken due to issue supplying invalid base64 byte dummy data.
```sh
E       binascii.Error: Invalid base64-encoded string: number of data characters (5) cannot be 1 more than a multiple of 4
```
https://github.com/HENNGE/aiodynamo/blob/fdb05e944fa2335017201dc42e446b45c0ea57df/benchmarks/deserialize/utils.py#L1-L16

Seems like this issue was fixed in unit tests but not propagated to benchmarks https://github.com/HENNGE/aiodynamo/blob/db16b2437751eb9755c7324be4e9230ff257eef1/tests/unit/test_utils.py#L31-L47

I propose we just copy the implementation over since there's no easy way to share test utils with benchmark because of the folder benchmarks are expected to run from.

stupoid:
Resolves #75

### DIFF
--- a/benchmarks/deserialize/utils.py
+++ b/benchmarks/deserialize/utils.py
@@ -1,7 +1,14 @@
+import base64
+
+
 def generate_item(nest):
     item = {
-        "hash": {"S": "string"},
-        "range": {"B": b"bytes"},
+        "hash": {
+            "S": "string",
+        },
+        "range": {
+            "B": base64.b64encode(b"bytes"),
+        },
         "null": {"NULL": True},
         "true": {"BOOL": True},
         "false": {"BOOL": False},
@@ -9,7 +16,9 @@ def generate_item(nest):
         "float": {"N": "4.2"},
         "numeric_set": {"NS": ["42", "4.2"]},
         "string_set": {"SS": ["hello", "world"]},
-        "binary_set": {"BS": [b"hello", b"world"]},
+        "binary_set": {
+            "BS": [base64.b64encode(b"hello"), base64.b64encode(b"world")]
+        },
     }
     if nest:
         item["list"] = {"L": [{"M": generate_item(False)}]}


benchmarks/deserialize/utils.py
```diff
@@ -1,15 +1,24 @@
...
+
 def generate_item(nest):
     item = {
-        "hash": {"S": "string"},
-        "range": {"B": b"bytes"},
+        "hash": {
+            "S": "string",
+        },
+        "range": {
+            "B": base64.b64encode(b"bytes"),
```
dimaqq, line 10: Alternatively, use these valid base-64 strings:

```
"bytesA=="
"helloA=="
"worldA=="
```

dimaqq (commented):
I'm not sure what the expected behaviour of `generate_item` is.
Is it meant to take in dynamo-style nested dict verbatim?
Is it meant to take in Python data and return a dynamo-styled nested dict?

What bugs me, both about the code before and after, is that strings and bytes are mixed in this structure.

It looks like a dynamo-styled thing, but such that would need to be json-able, and therefore should not have `bytes` in it.

ojii:
> I'm not sure what the expected behaviour of `generate_item` is.
> Is it meant to take in dynamo-style nested dict verbatim?
> Is it meant to take in Python data and return a dynamo-styled nested dict?
> 
> What bugs me, both about the code before and after, is that strings and bytes are mixed in this structure.
> 
> It looks like a dynamo-styled thing, but such that would need to be json-able, and therefore should not have `bytes` in it.

It's supposed to be a DynamoItem as it gets returned from `json.loads`. 

I think what happened here is a holdover from very early aiodynamo because of how [boto3 works](https://github.com/boto/boto3/blob/develop/boto3/dynamodb/types.py) where the deserializer already gets bytes, not b64 strings (their json decoder seems to be base64 aware or something like that, not sure what's going on there). The benchmarks just never got updated to reflect those bytes/base64 fixes.

ojii (approved):

PR #78, opened 2021-07-03
Update deserialize to use mapping

### LINKED ISSUE #74: Speed up deserialization
When dealing with large items/large number of items, [`deserialize`](https://github.com/HENNGE/aiodynamo/blob/fdb05e944fa2335017201dc42e446b45c0ea57df/src/aiodynamo/utils.py#L25-L53) quickly becomes a bottleneck. The current version is already an optimized version over what botocore does (which is a major reason why aiodynamo is faster than botocore), but there is probably still room for improvement.

One option would be to include an (optional) cython version of the function.

stupoid:
Resolves #74

### DIFF
--- a/src/aiodynamo/utils.py
+++ b/src/aiodynamo/utils.py
@@ -4,9 +4,12 @@ import decimal
 import logging
 from collections import abc as collections_abc
 from functools import reduce
-from typing import Any, Callable, Dict, Mapping, Tuple, Union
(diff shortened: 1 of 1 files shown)


dimaqq (approved):
🎉

dimaqq:
@ojii how about we merge this and release?

ojii (approved):

ojii:
🤔 why did circleci not run?

PR #86, opened 2021-08-24
Count limits

Tinche:
This was motivated by our real-life use case. We sometimes need to limit counting to preserve latency and Dynamo resources. I tried using `query_single_page` but it doesn't support count queries, so I just added `limit` to `count`.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+UNRELEASED
+----------
+
+Release Date: UNRELEASED
+
+* Added the `limit` parameter to :py:meth:`aiodynamo.client.Client.count`
+
 21.8
 ----
 
--- a/src/aiodynamo/client.py
+++ b/src/aiodynamo/client.py
@@ -923,7 +923,8 @@ class Client:
                     task = None
                 else:
                     if limit is not None:
-                        limit = limit - res["Count"] if is_count else len(res["Items"])
+                        consumed: int = res["Count"] if is_count else len(res["Items"])
+                        limit -= consumed
                         if limit > 0:
                             payload["Limit"] = limit
                         else:
--- a/src/aiodynamo/client.py
+++ b/src/aiodynamo/client.py
@@ -271,6 +271,7 @@ class Table:
         start_key: Optional[Dict[str, Any]] = None,
         filter_expression: Optional[Condition] = None,
         index: Optional[str] = None,
+        limit: Optional[int] = None,
     ) -> int:
         return await self.client.count(
             self.name,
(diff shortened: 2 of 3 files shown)


Tinche:
The mypy failure is at `src/aiodynamo/http/httpx.py:45: error: Argument "data" to "post" of "AsyncClient" has incompatible type "bytes"; expected "Dict[Any, Any]"` which I haven't touched. Should probably be fixed separately.

docs/changelog.rst
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+UNRELEASED
```
dimaqq, line 4: TIL where the change log is in this project :orz:

dimaqq:
> The mypy failure is at `src/aiodynamo/http/httpx.py:45: error: Argument "data" to "post" of "AsyncClient" has incompatible type "bytes"; expected "Dict[Any, Any]"` which I haven't touched. Should probably be fixed separately.

Will be done in #87

src/aiodynamo/client.py
```diff
@@ -907,19 +910,21 @@ async def _depaginate(
         task: Optional[asyncio.Task[Dict[str, Any]]] = asyncio.create_task(
             self.send_request(action=action, payload=payload)
         )
+        is_count = payload.get("Select") == "COUNT"
         try:
             while task:
-                result = await task
+                res = await task
```
dimaqq, line 916: Out of curiosity, what's the rationale for `s/result/res/` ?

tests/integration/test_client.py
```diff
@@ -90,6 +90,7 @@ async def test_count(client: Client, table: TableName):
         await client.count(table, HashKey("h", "h1") & RangeKey("r").begins_with("x"))
         == 0
     )
+    await client.count(table, HashKey("h", "h2"), limit=1) == 1
```
dimaqq, line 93: I'd say, make it a separate test 🤔

src/aiodynamo/client.py
Tinche: Stops black from breaking up the line below when we use it

tests/integration/test_client.py
Tinche: Sure, can do tomorrow

Tinche: Done!

src/aiodynamo/client.py
```diff
@@ -745,6 +747,7 @@ async def count(
         start_key: Optional[Dict[str, Any]] = None,
         filter_expression: Optional[Condition] = None,
         index: Optional[str] = None,
+        limit: Optional[int] = None,
```
dimaqq, line 750: Documentation needs to be updated :)

dimaqq (approved):
maybe I'll slop limit onto the docs after merge :clever:

PR #94, opened 2021-09-16
Correct "decorrelated" spelling

dimaqq:


ojii (approved):
"luckily" `ThrottleConfig` isn't documented (it should be!) so I think we can get by with saying this isn't a breaking change.

PR #98, opened 2021-10-08
Implement scan_count method

TheEdgeOfRage:
Same as count, but uses scan instead of query.


ojii:
thank you.

TheEdgeOfRage:
@ojii Could you please add the `hacktoberfest-accepted` label here as well. Thanks <3

PR #101, opened 2021-10-20
Support consistent read

jarikujansuu:
## Purpose

Support consistent read in get_item, query, scan, count and batch get operations.

**Note**: Tests do not actually check anything except that DynamoDb calls do not explode if using `consistent_read` option. Local DynamoDb is always strongly consistent. And might want to reduce copied code in them.

### DIFF
--- a/src/aiodynamo/client.py
+++ b/src/aiodynamo/client.py
@@ -135,14 +135,20 @@ class Table:
         )
 
     async def get_item(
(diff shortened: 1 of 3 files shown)


ojii:
> Tests do not actually check anything except that DynamoDb calls do not explode if using `consistent_read` option.

that's fine, but could you please change the tests to use `@pytest.mark.parametrize` since the new tests seem to just be a copy/paste of existing ones with a single parameter changed.

PR #106, opened 2021-11-16
automate pypi deploy

### LINKED ISSUE #105: Faster release cycle
Not sure if this is correct channel for this kind of request. 

But I wish there was faster release cycle, currently of course interested in having released version which would include https://github.com/HENNGE/aiodynamo/pull/101. Last release has been October 7th and after that has been 2 functional improvement PRs, and that consistent read support was merged 23 days ago.

I am now using git dependency but as there are also no tags have to rely on random rev value which is not obviously clear what we are using.

tedchou12:
closes #105 
**Need to add `PYPI_TOKEN` to secret**.


ojii:
it shouldn't release without tests passing, so this should either be moved to cci or the cci stuff needs to be moved to GHA. 

Also, doesn't this need some sort of "is this actually a new version" check beyond just "it's on the master branch"?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+      uses: actions/checkout@v2
+
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+
+    - name: Install packages
+      run: |
+        sudo apt-get install libpq-dev -y
```
dimaqq, line 27: Postgres? Why is that needed?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+        python3 -m pip install wheel twine setuptools poetry
```
dimaqq, line 28: I don’t know for sure, but I think twine is alternative to poetry… it’s odd that both are required

dimaqq: While on the subject, I think that poetry should be self-sufficient 🤔 
I can’t be totally sure, but I would not be surprised if wheel and setup tools are redundant here

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: Configure poetry
+      run: |
+        poetry config pypi-token.pypi ${{ secrets.PYPI_TOKEN }}
+
+
+    - name: pypi deployments
+      run: |
+        poetry build
```
dimaqq, line 37: Does this run after or in parallel to the unit tests?

dimaqq:
> it shouldn't release without tests passing, so this should either be moved to cci or the cci stuff needs to be moved to GHA.
> 
> Also, doesn't this need some sort of "is this actually a new version" check beyond just "it's on the master branch"?

pypi will reject reupload of the same version.
thus, it's kinda :fine:
(not totally fine, because the action needs to silence upload errors)
(fine here because it's one package for all arch's, but would not be fine if we built arch-specific wheels)

I think perhaps the GHA way would be to trigger on release rather than on push?

If we wanted to hack this manually:
Maybe it's possible to examine the last commit diff to see if the version changed?
Or poke pypi and see if the local version is different?

.github/workflows/deploy.yml
tedchou12: sorry, those are not needed.

.github/workflows/deploy.yml
tedchou12: I tried running without wheel, twin and setuptools. Yes, seems like poetry is sufficient.

.github/workflows/deploy.yml
tedchou12: As @ojii suggested, putting to circleci workflow is probably better!

tedchou12:
yes, I think moving to CircleCI is better. I made the tests required before deploy.
> Also, doesn't this need some sort of "is this actually a new version" check beyond just "it's on the master branch"?
As @dimaqq suggested, the poetry will automatically reject if a same version is uploaded with 400 error, so I guess it will hint the necessity to bump the version.
Is possible to compare the versions before upload, but I guess just doing a overall checksum is not safe while getting the version number is a bit messy.
`import toml;pkg_v = toml.load('pyproject.toml')['tool']['poetry']['version'];import luddite;pypi_v=luddite.get_version_pypi('aiodynamo')\nif (pypi_v != pkg_v):  print('1')\nelse:  print('0');`

to determine which branch should trigger the action, is possible to specify with the following filter, sorry, I am not sure about the workflow for this, I will leave this part to you.

```
filters:
            branches:
              only:
                - master
```

As for the `${PYPI_TOKEN}`, it is needed to be inputted into the environmental variables in circleci.

dimaqq:
This effort needs a new champion!

ojii:
i've tried to port cci to gha but got disillusioned by [act](https://github.com/nektos/act) https://github.com/HENNGE/aiodynamo/tree/gha

dimaqq:
Hmm unhashable VersionUnion was a poetry bug https://github.com/python-poetry/poetry/issues/2340
I wonder what poetry version the action you use pulls in?

dimaqq:
My 2c: how about smaller scope: for example, only py3.9, only `pytest` tests; leave mypy/isort/versions to the maintainers, at least at the start?

dimaqq:
re: poetry failing, it's possible to specify exactly what poetry version to use:

```yml
        uses: abatilo/actions-poetry@v2.0.0
        with:
          poetry-version: ${{ matrix.poetry-version }}
```

(perhaps a constant in this case)

ojii:
> re: poetry failing, it's possible to specify exactly what poetry version to use:

it's supposed to use 1.1.11 https://github.com/abatilo/actions-poetry/blob/7044c9c69e0265717d52471f66033b8d0e2a69ff/action.yml

ojii:
closing this since it's pretty out of date now due to #122

This should probably be discussed a bit more in #105, specifically with regards to _when_ to publish a new version and how/where version numbers are handled.

PR #118, opened 2022-02-07
Fix new-http error handling

ojii:
On errors (eg throttling), no exception would be raised, causing higher level APIs to raise key errors.


dimaqq (commented):
Nice!

PR #120, opened 2022-02-08
Fix mypy type hints

### LINKED ISSUE #119: Incompatible argument type issues after version 22.2
Mypy has started to give errors such as `Client" has incompatible type "HTTPX"; expected "FixedCallable[Request, Awaitable[Response]]` after version 22.2. This happens when trying to use aiodynamo as documented here https://github.com/HENNGE/aiodynamo#with-httpx and running mypy (eg with default settings).

This also happens in the repository it self when running mypy over the unit test code. Same errors happen eg in https://github.com/HENNGE/aiodynamo/blob/master/tests/integration/conftest.py#L142. As an idea it would be good to run mypy over the unit test code here instead of just running it over `src` as defined in the `mypy.ini`. It would avoid the type breakage:

```
aiodynamo % poetry run mypy tests
... some errors omitted ...
tests/unit/test_errors.py:26: error: Argument 1 to "Client" has incompatible type "AIOHTTP"; expected "FixedCallable[Request, Awaitable[Response]]"
tests/unit/test_errors.py:51: error: Argument 1 to "Client" has incompatible type "AIOHTTP"; expected "FixedCallable[Request, Awaitable[Response]]"
tests/unit/test_client.py:43: error: Argument 1 to "Client" has incompatible type "Callable[[Request], Coroutine[Any, Any, Response]]"; expected "FixedCallable[Request, Awaitable[Response]]"
tests/integration/conftest.py:114: error: Incompatible return value type (got "Callable[[Throughput], Coroutine[Any, Any, str]]", expected "Callable[[Optional[Throughput]], Awaitable[str]]")
tests/integration/conftest.py:142: error: Argument 1 to "Client" has incompatible type "HTTPX"; expected "FixedCallable[Request, Awaitable[Response]]"
tests/integration/conftest.py:160: error: Argument 1 to "Client" has incompatible type "HTTPX"; expected "FixedCallable[Request, Awaitable[Response]]"
Found 52 errors in 8 files (checked 11 source files)
```

ojii:
Fixes #119 

Still can't use `Callable` with dataclasses, but looks like my attempt to make a generic `Protocol` to fix the issue failed, so now just using protocols directly. 

Also added tests to mypy coverage to catch such issues in the future, even though typing tests is rather painful.


tests/unit/test_credentials.py
```diff
@@ -1,10 +1,13 @@
 import datetime
 from pathlib import Path
 from textwrap import dedent
+from typing import AsyncGenerator, Optional
 
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
 from aiohttp import web
 from freezegun import freeze_time
+from pyfakefs.fake_filesystem import FakeFilesystem  # type: ignore
```
MarkusSintonen, line 10: I usually prefer using explicit ignores so specifying the failure kind: `# type: ignore[import]`. Same comment for other places with ignores here. But this is not too important to fix as it depends on the conventions of the codebase

MarkusSintonen (approved):
Awesome, thank you! 💯 

Also verified locally that the reported issue doesn't happen anymore:
```
aiodynamo % poetry run mypy
Success: no issues found in 25 source files
aiodynamo % poetry run mypy tests
Success: no issues found in 11 source files
```

dimaqq (approved):
Nice, it's much cleaner with a Protocol

dimaqq (approved):

tests/unit/test_credentials.py
ojii: wasn't aware of that feature in mypy, thanks for sharing. updated all the ignore comments to ignore a specific error.

PR #122, opened 2022-02-16
circleci ➡️ github actions

ojii:
would be nice to reduce some of the repetition but i'm just happy it works at all.


ojii:
by far the slowest part of this whole ci pipeline is poetry install without a lock file. I wonder if it's possible to only run this once (or rather, run `poetry lock`) and then re-use it in subsequent steps.

dimaqq (approved):
Merge it, improve later!

dimaqq:
> by far the slowest part of this whole ci pipeline is poetry install without a lock file. I wonder if it's possible to only run this once (or rather, run `poetry lock`) and then re-use it in subsequent steps.

maybe that's a good reason to check in `poetry.lock`, finally?

ojii:
> maybe that's a good reason to check in `poetry.lock`, finally?

I still think that's a bad practice for _libraries_.

ojii:
> Merge it, improve later!

i wanna mess in this branch a bit until it's somewhat decent and then squash the commits

PR #127, opened 2022-06-28
Add kubernetes credentials to the list of possible credentials

thiagosalvatore:
We've been using AioDynamo for a while in my current company. The existing credentials work pretty well with ECS and EC2 machines. However, when running inside a Kubernetes cluster (like EKS), the metadata from the EC2 instance is not available and we don't have the credentials from AWS as environment variables because our pod basically assumes a role on AWS. We also don't have a file with credentials or anything like that. That said, we needed a way to fetch the credentials from AWS that our pod has, and luckily AWS offers us a way to do that by calling `get_credentials`.

This PR does the following:

1. Creates a new `KubernetesCredentials`;
2. Updates typing-extensions to support the most recent version;
3. Add boto3-stubs to our dev dependencies;

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ python = "^3.7"
 httpx = {version = ">=0.15.0 <1.0.0", optional = true}
 aiohttp = {version = "^3.6.2", optional = true}
 yarl = "^1.4.2"
-typing_extensions = { version = "^3.7", python = "< 3.8" }
+typing_extensions = { version = "^4.2", python = "< 3.8" }
 
 [tool.poetry.extras]
 httpx = ["httpx"]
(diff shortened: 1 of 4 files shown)


ojii (changes requested):
Thank you for your pull request.

Unfortunately, there's a few issues with it. Using `boto3` results in blocking calls being made which is not suitable for an async library. From what I understand reading the [documentation](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html) on how credentials work in EKS, what is needed to support it in aiodynamo is implementing `AssumeRoleWithWebIdentity`. Supporting that "natively" would be nice, but would need a much stronger test suite and not rely on mocks that much.

Also, please try to keep the code similar to the rest of the code, eg not using `self.__...` and using the existing logging infrastructure.

ojii:
I think I know how to implement this natively (we have a custom credentials implementation for connecting with a saml response which is very similar) but my biggest concern is how to test this on EKS, since I've never touched either k8s or EKS. could you provide some guide how I could test the implementation?

ojii:
I'm sorry, but I will close this PR. I'd love to add support for this, but the first step to do so would be to provide a way to test it and the implementation needs to be async native to be included in this repo.

PR #130, opened 2022-07-11
Fix integration tests not running on Github Actions

### LINKED ISSUE #129: after moving to github actions, integration tests are no longer run
Need to figure out how to run dynalite in github actions. or ideally run against both dynalite and dynamodb local.

hanscg:
Closes #129 

Integration tests now run with both dynalite and dynamodb-local.
Unfortunately service can't be set with matrix so the code has to be duplicated

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -27,7 +27,7 @@ jobs:
     runs-on: "ubuntu-latest"
     services:
       dynalite:
-        image: dimaqq/dynalite:latest
+        image: ghcr.io/dimaqq/dynalite:latest
         ports:
           - 8000:8000
     env:
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           path: poetry.lock
           key: ${{ github.sha }}-${{ matrix.python-version }}
-  pytest:
+  pytest-dynalite:
     needs: lock
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yml
```diff
@@ -19,12 +19,48 @@ jobs:
...
+    needs: lock
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
+    runs-on: "ubuntu-latest"
+    services:
+      dynamodb:
+        image: amazon/dynamodb-local:latest
+        ports:
+          - 8000:8000
```
dimaqq, line 59: Ugh that's a bummer that services don't grok matrix....
I guess there's another alternative:
* 1 job
* has 2 services on 2 ports
* matrix which port is used by `pytest`

Perhaps that's too clever though -- hard to tell what fails if it's `port 8001` 🤔

dimaqq (approved):

.github/workflows/workflow.yml
```diff
@@ -19,12 +19,48 @@ jobs:
...
-  pytest:
+  pytest-dynalite:
     needs: lock
     strategy:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
     runs-on: "ubuntu-latest"
+    services:
+      dynalite:
+        image: dimaqq/dynalite:latest
```
dimaqq, line 30: Hey folks, I've moved the image to ghcr.io though it's documented poorly 😅 
https://github.com/dimaqq/dynalite-docker

dimaqq (approved):
Awesome, Hans!

ojii (approved):
thank you

Issue #53, opened 2020-05-26
question: Any chance for transaction support?


stupoid:
I'm wondering if there are any plans to support [TransactWriteItems](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html) or just [Transactions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transactions.html) in general.

I can probably work on this if/when I have more time 🤔

ojii:
No plans to add support right now, but would be nice to have of course.

sadly, transactions seem to be unsupported by dynalite (or alternative dynamodb emulators) which will make testing this difficult (basically requiring "real" dynamodb). 

We should probably discuss API design before you dive too deep. At a first glance, I think `TransactItems` should probably built similar to `UpdateExpression`, using `&` to combine different operations (put, update,...) into a single `TransactItems` object.

Maybe the signature should be `transact_write_items(self, table: TableName, items: TransactItems, *, request_token: Optional[str] = None) -> None:`, what do you think?

dimaqq:
@stupoid do you use either transaction mechanism already?
if not, please read up on the limitations, they are hilarious :)

dimaqq:
Calling on users who want transactions... please let yourselves known, e.g. by 👍🏿  on the OP ⬆️

nicolaszein:
Hi @ojii, how are you?

I'm actually also interested in transaction support.

If you are still interested, I have the time to collaborate with this. 😄 
About the dynalite not supporting transactions I have cloned the repo and ran the test suit against a [localstack](https://hub.docker.com/r/localstack/localstack) dynamodb and all the tests succeed. So I think it is an option what do you think?

And about the API design, I liked the idea of building similar to `UpdateExpression`. But I have some doubts about how to implement it. `UpdateExpression` is a collection of `F` expressions right? Do you think that `TransactItems` should be a collection of `TransactItem`?

I just want to hear your thoughts about and if possible collaborate with this. 😃

ojii:
> About the dynalite not supporting transactions I have cloned the repo and ran the test suit against a [localstack](https://hub.docker.com/r/localstack/localstack) dynamodb and all the tests succeed. So I think it is an option what do you think?

[...]

PR #131, opened 2022-08-02
Add localstack pytest step on CI workflow

nicolaszein:
As mentioned on #53 issue, to be able to support dynamo transaction we need another dynamo emulator because dynalite does not support transaction.

So this PR is adding a step on CI that run tests against a localstack dynamodb.


hanscg (approved):
LGTM, thanks! 👍

ojii:
Thank you @nicolaszein

PR #132, opened 2022-08-05
Add transact write items method to client

nicolaszein:
Adding support for [TransactWriteItems](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html)

### DIFF
--- a/src/aiodynamo/client.py
+++ b/src/aiodynamo/client.py
@@ -4,9 +4,21 @@ import asyncio
 import datetime
 import json
 from dataclasses import dataclass
-from typing import Any, AsyncIterator, Dict, List, Mapping, Optional, Union, cast
+from typing import (
+    Any,
+    AsyncIterator,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Union,
+    cast,
+)
 
 from yarl import URL
+
+from aiodynamo.operations import ConditionCheck, Delete, Put, Update
 
 from .credentials import Credentials
 from .errors import (
@@ -23,6 +35,8 @@ from .errors import (
     TableNotFound,
     Throttled,
     TimeToLiveStatusNotChanged,
+    TooManyTransactions,
+    TransactionEmpty,
     exception_from_response,
 )
 from .expressions import (
@@ -929,6 +943,25 @@ class Client:
                     )
             result[table] = BatchWriteResult(undeleted_keys, unput_items)
         return result
+
+    async def transact_write_items(
+        self,
+        items: Sequence[Union[Put, Update, Delete, ConditionCheck]],
+        *,
+        request_token: Optional[str] = None,
+    ) -> None:
+        if len(items) == 0:
+            raise TransactionEmpty("TransactWriteItems must have at least 1 operation")
+        if len(items) > 25:
+            raise TooManyTransactions(
+                "TransactWriteItems must have a maximum of 25 operations"
+            )
+
+        payload = {
+            "ClientRequestToken": request_token,
+            "TransactItems": [item.to_request_payload() for item in items],
+        }
+        await self.send_request(action="TransactWriteItems", payload=payload)
 
     async def send_request(
         self,
--- a/src/aiodynamo/errors.py
+++ b/src/aiodynamo/errors.py
@@ -1,4 +1,5 @@
 import json
+from typing import Any, Dict
 
 
 class AIODynamoError(Exception):
@@ -112,6 +113,14 @@ class TransactionCanceled(AIODynamoError):
     pass
 
 
+class TransactionEmpty(AIODynamoError):
+    pass
+
+
+class TooManyTransactions(AIODynamoError):
+    pass
+
+
 class ReplicaAlreadyExists(AIODynamoError):
     pass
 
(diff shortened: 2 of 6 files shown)


src/aiodynamo/client.py
```diff
@@ -930,6 +943,27 @@ async def batch_write(
...
+        request_token: Optional[str] = None,
+    ) -> None:
+        if len(items) == 0:
+            raise TransactionCanceled(
+                "TransactWriteItems must have at least 1 operation"
+            )
+        if len(items) > 25:
+            raise TransactionCanceled(
+                "TransactWriteItems must have a maximum of 25 operations"
+            )
```
ojii, line 959: I would prefer two separate exceptions for these cases which are distinct so you don't need to check the string to figure out what went wrong. Feel free to come up with better names but maybe something like `TransactionEmpty` and `TransactionTooLarge`

tests/integration/conftest.py
```diff
@@ -49,6 +49,11 @@ def real_dynamo() -> bool:
     return os.environ.get("TEST_ON_AWS", "false") == "true"
 
 
+@pytest.fixture(scope="session")
+def is_dynalite_emulator() -> bool:
+    return os.environ.get("DYNALITE_EMULATOR", "false") == "true"
```
ojii, line 54: I think it would be better to do feature detection rather than relying on the person/system running the system setting an env var. Maybe something like this:

```python
@pytest.fixture()
async def supports_transactions(client: Client, table: TableName):
    try:
        await client.transact_write_items(
            [ConditionCheck(table, {"h": "h", "r": "r"}, F("h").does_not_exist())]
        )
    except UnknownOperation:
        raise pytest.skip(f"Transactions not supported")
```

Then you can `@pytest.mark.usefixtures('supports_transactions')` the tests to skip on implementations that don't support it.

The benefit of this is also that if dynalite does eventually add support, we don't need to change anything, or if we add yet another implementation to our test suite we don't have to worry about it either.

tests/integration/test_client.py
```diff
@@ -589,6 +590,124 @@ async def test_batch(client: Client, table: TableName) -> None:
...
+) -> None:
+    if is_dynalite_emulator:
+        pytest.xfail("transaction not supported by dynalite yet")
+
+    with pytest.raises(errors.TransactionCanceled):
+        await client.transact_write_items(items=[])
+
+    with pytest.raises(errors.TransactionCanceled):
+        puts = [Put(table=table, item={"h": "h", "r": str(i)}) for i in range(26)]
+        await client.transact_write_items(items=puts)
```
ojii, line 604: I think this test shows that we need to handle `TransactionCanceled` in a better way. The [documentation](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html#API_TransactWriteItems_Errors) shows that that error has a nested error which hints at the actual issue. I think we should capture that inner error and raise that instead.

tests/integration/test_client.py
```diff
@@ -589,6 +590,124 @@ async def test_batch(client: Client, table: TableName) -> None:
...
+    deletes = [
+        Delete(
+            table=table,
+            key={"h": "h", "r": str(i)},
+        )
+        for i in range(2)
+    ]
+    response = await client.transact_write_items(items=deletes)
+    assert not response
+    assert len([item async for item in client.query(table, HashKey("h", "h"))]) == 0
```
ojii, line 635: It's great to see all these tests, but could you please split them up into individual tests all testing a specific part of the API? This test for example checks input validation (which could be a single test function with `pytest.mark.parametrize` for the two cases) as well as the `Put`, `Update` and `Delete` operations which should also be their own test.

The shorter the test the easier it is to see what is tested and also failures in one part of the test don't hide failures in later parts.

ojii (changes requested):
Thank you for implementing this. The pull request looks mostly good, but there's a few things I'd like you to change.

nicolaszein:
Hi @ojii, thanks for your review.

I have implemented the suggestions that you said.

Can you please take a look again? 😃

src/aiodynamo/errors.py
```diff
@@ -188,6 +197,17 @@ def exception_from_response(status: int, body: bytes) -> Exception:
...
     except Exception:
         return UnknownError(status, body)
+
+
+def extract_error_from_transaction_canceled(data: Dict[str, Any]) -> AIODynamoError:
+    try:
+        error = data["CancellationReasons"][0]
+        return ERRORS[f"{error['Code']}Exception"](error["Message"])
+    except Exception:
+        return ERRORS[data["__type"].split("#", 1)[1]](data)
```
ojii, line 213: `IdempotentParameterMismatchException` and `TransactionInProgressException` are not yet in `ERRORS` but can happen according to the docs

nicolaszein: Nice catch! Just added these exceptions 😃

ojii (commented):
looking good. Please add the new API and the types it uses to the [docs](https://github.com/HENNGE/aiodynamo/blob/master/docs/usage.rst)

nicolaszein:
> looking good. Please add the new API and the types it uses to the [docs](https://github.com/HENNGE/aiodynamo/blob/master/docs/usage.rst)

Nice! Just added the documentation! 😃

ojii:
Thank you very much for implementing it. Looks like I messed up with git a little bit when I merged this because I wanted to fix a small rst issue before merging and now Github doesn't detect this as merged, so I will instead close it (but it's actually merged from what I can see).

Do you also plan to add `TransactGetItems`?

nicolaszein:
Thank you @ojii for all your support! :)

I already opened the #135 for transact get items.

PR #136, opened 2022-09-08
Update mypy and remove mypy_hack

akawasaki:
### Description
Updated `mypy` to latest version to revert some hacks done for #120 

- Updated `mypy` to `0.971`
- revert mypy hacks done for #120 
- Resolve mypy error `Function does not return a value` 
- Some tests improvements


src/aiodynamo/client.py
```diff
@@ -949,7 +949,7 @@ async def transact_write_items(
         items: Sequence[Union[Put, Update, Delete, ConditionCheck]],
         *,
         request_token: Optional[str] = None,
-    ) -> None:
+    ) -> Any:
```
dimaqq, line 952: This function never explicitly returns. Why `Any`?

tests/integration/test_client.py
```diff
@@ -618,7 +620,7 @@ async def test_transact_write_items_put(client: Client, table: TableName) -> Non
         Put(table=table, item={"h": "h", "r": str(i), "s": "initial"}) for i in range(2)
     ]
     response = await client.transact_write_items(items=puts)
-    assert not response
+    assert response is None
```
dimaqq, line 623: IMO, `not X` is more Pythonic.
In this case, however, since `transact_xxx` never returns anything, the assertion is best removed.

tests/unit/test_client.py
```diff
@@ -1,5 +1,5 @@
 import json
-from typing import Any, Type
+from typing import Any, Type, cast
```
dimaqq, line 2: `cast` import doesn't seem to be used in this PR

tests/unit/test_errors.py
```diff
@@ -9,7 +9,7 @@
 from aiodynamo.credentials import Key, StaticCredentials
 from aiodynamo.errors import ValidationException
 from aiodynamo.http.aiohttp import AIOHTTP
-from aiodynamo.http.types import Request, Response
+from aiodynamo.http.types import HttpImplementation, Request, Response
```
dimaqq, line 12: the added import doesn't appear to be used in this PR

tests/unit/test_errors.py
```diff
@@ -72,6 +72,11 @@ async def test_dynamo_retries_50x(status: int) -> None:
     async def http(request: Request) -> Response:
         return next(responses)
 
-    client = Client(http, creds, "test", throttle_config=StaticDelayRetry(delay=0.01))
+    client = Client(
+        http,
+        creds,
+        "test",
+        throttle_config=StaticDelayRetry(delay=0.01),
+    )
```
dimaqq, line 80: This format change is odd, because original line was <88 chars.
Do you have `pre-commit` installed in this repo?

tests/integration/conftest.py
```diff
@@ -154,7 +154,12 @@ def prefilled_table(
 
     async def startup() -> str:
         async with AsyncClient() as session:
-            client = Client(HTTPX(session), Credentials.auto(), region, endpoint)
+            client = Client(
+                HTTPX(session),
+                Credentials.auto(),
+                region,
+                endpoint,
+            )
```
dimaqq, line 162: this formatting change is odd too.

dimaqq (commented):
I'm very glad to see that `mypy` got better and that typing hacks can be removed.

There are a few changes in the PR that seem unrelated, like imports and style

dimaqq:
> Resolve mypy error `Function does not return a value`

Apparently there's a flag for that, see https://github.com/python/mypy/pull/13219

src/aiodynamo/client.py
akawasaki: Yeah to resolve mypy error `function does not return value` and it seems the default return type for a missing annotation is `Any` in this case?

akawasaki: Ah okay, will take a look at https://github.com/python/mypy/pull/13219 🙏

tests/unit/test_errors.py
akawasaki: yes i do have `pre-commit` installed 🤔 will dig in

src/aiodynamo/http/types.py
```diff
@@ -23,6 +23,4 @@ class RequestFailed(Exception):
     inner: Exception
 
 
-class HttpImplementation(Protocol):
-    async def __call__(self, request: Request) -> Response:
-        ...
+HttpImplementation = Callable[[Request], Awaitable[Response]]
```
dimaqq, line 26: `Protocol` import can be removed

dimaqq: and maybe if all the uses of Protocol are done, it can be removed from `_compat`

dimaqq (approved):

tests/integration/conftest.py
```diff
@@ -35,9 +23,11 @@
     StaticDelayRetry,
     Throughput,
 )
+from aiodynamo.operations import ConditionCheck
+from aiodynamo.types import TableName
 
 
-class TableFactory(Protocol):
+class TableFactory:
```
ojii, line 30: this can also become a Callable, right?

dimaqq (approved):

ojii (approved):
Thank you for these improvements. So happy these ugly hacks could finally be removed

PR #137, opened 2022-09-20
Support typed exceptions with dynamodb-local.

aclemons:
The value of the __type key in body of the error responses from dynamodb local does not have the com.amazonaws.dynamodb.v20120810# prefix.

An example of a ConditionalCheckFailedException:

```
{"__type": "ConditionalCheckFailedException", "message": "The conditional request failed"}
````


ojii:
weird behavior of dynamodb-local here... anyway, thanks for the PR, looks good.

aclemons:
I also could not find any documentation of why the error values are different, which was frustrating.

PR #141, opened 2022-10-25
Add test for update with IN condition

ojii:
I think I found a bug in dynalite, so for now I just have a test until I have more information. 

EDIT: slight scope creep, added compatibility for pytest-xdist because on my machine running the full test suite on three implementations took longer than one minute, with `-nauto` it's down to 30 seconds again.


ojii:
It's a dynalite bug https://github.com/mhart/dynalite/pull/159

ojii:
tested against real dynamodb and passes

dimaqq (approved):

Issue #142, opened 2022-10-28
Support AWS Web Identity Token File authentication


mrkovalchuk:
`aiodynamo` doesn't support AWS Web Identity Token File authentication. 

It looks like a typical case. For my team, we would like to use that authentication in our services in k8s by using [IAM Roles for Service Accounts](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)(IRSA)

Related issue: https://github.com/HENNGE/aiodynamo/issues/128
Related PR: https://github.com/HENNGE/aiodynamo/pull/127

There was a similar issue, but I decided to create another one because I think we can describe our goal and issue more directly. 

So, as discovered in a related PR, we should implement the `AssumeRoleWithWebIdentity` API for this authentication.

ojii:
> `aiodynamo` doesn't support AWS Web Identity Token File authentication.

sorry to be nit-picky, but `aiodynamo` technically supports any kind of authentication you want, because it allows you to provide a [custom credentials loader](https://aiodynamo.readthedocs.io/en/latest/advanced.html#custom-credentials-loader).

I assume you meant to say "doesn't support it out of the box" and that's fair.

As in #128 #127, my biggest concern with adding this to aiodynamo itself is "how do we test this" (both in unit tests/integration tests and manually on actual EKS). I do not use k8s so if you want to provide an implementation of this credentials loader, please make sure to avoid the issues outlined in #127 and provide clear and simple instructions on how to verify your implementation on EKS/k8s.

mrkovalchuk:
> sorry to be nit-picky, but aiodynamo technically supports any kind of authentication you want, because it allows you to provide a [custom credentials loader](https://aiodynamo.readthedocs.io/en/latest/advanced.html#custom-credentials-loader).

You're right =) Of course, I meant "doesn't support it out of the box", sorry about it 👍 
Also, I didn't expect your answer so quickly!

I prepare some workaround that works for me. I need approval on a concept.

> and provide clear and simple instructions on how to verify your implementation on EKS/k8s.
👍

PR #143, opened 2022-10-28
Add support for AWS Web Identity Token File authentication

mrkovalchuk:
Add support for AWS Web Identity Token File authentication.

Related issue: https://github.com/HENNGE/aiodynamo/issues/142

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ python = "^3.7"
 httpx = {version = ">=0.15.0 <1.0.0", optional = true}
 aiohttp = {version = "^3.6.2", optional = true}
 yarl = "^1.4.2"
+defusedxml = "^0.7.1"
 typing_extensions = { version = "^3.7", python = "< 3.8" }
 
 [tool.poetry.extras]
--- a/src/aiodynamo/credentials.py
+++ b/src/aiodynamo/credentials.py
@@ -6,11 +6,13 @@ import configparser
 import datetime
 import json
 import os
+import time
 from dataclasses import dataclass, field
 from enum import Enum, auto
 from pathlib import Path
-from typing import Callable, Optional, Sequence, TypeVar
+from typing import Any, Callable, Dict, Optional, Sequence, TypeVar
 
+from defusedxml import ElementTree as ET
 from yarl import URL
 
 from .http.types import HttpImplementation, Request, RequestFailed
@@ -34,6 +36,7 @@ class Credentials(metaclass=abc.ABCMeta):
         return ChainCredentials(
             candidates=[
                 EnvironmentCredentials(),
+                AssumeRoleWithWebIdentityCredentials(),
                 FileCredentials(),
                 ContainerMetadataCredentials(),
                 InstanceMetadataCredentials(),
@@ -287,6 +290,107 @@ def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
     if thing is not None:
         return mapper(thing)
     return thing
+
+
+AWS_STS_API_VERSION = "2011-06-15"
+
+# https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html
+@dataclass
+class AssumeRoleWithWebIdentityCredentials(MetadataCredentials):
+    ACTION = "AssumeRoleWithWebIdentity"
+
+    timeout: Timeout = 2
+    max_attempts: int = 3
+
+    base_url: URL = URL("https://sts.amazonaws.com/")
+
+    token_path: Optional[str] = field(
+        default_factory=lambda: os.environ.get("AWS_WEB_IDENTITY_TOKEN_FILE", None)
+    )
+    role_arn: Optional[str] = field(
+        default_factory=lambda: os.environ.get("AWS_ROLE_ARN", None)
+    )
+    role_session_name: str = field(
+        default_factory=lambda: os.environ.get(
+            "AWS_ROLE_SESSION_NAME", f"default-session-{time.time()}"
+        )
+    )
+
+    def __post_init__(self) -> None:
+        self._identity_token: Optional[str] = None
+        super().__post_init__()
+
+    async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
+        url = self._build_url()
+        response: bytes = await fetch_with_retry_and_timeout(
+            http=http,
+            max_attempts=self.max_attempts,
+            timeout=self.timeout,
+            request=Request(method="GET", url=url, headers=None, body=None),
+        )
+        credentials = self._take_credentials_from_response(response)
+        return Metadata(
+            key=Key(
+                id=credentials["AccessKeyId"],
+                secret=credentials["SecretAccessKey"],
+                token=credentials.get("SessionToken", None),
+            ),
+            expires=parse_amazon_timestamp(credentials["Expiration"]),
+        )
+
+    def is_disabled(self) -> bool:
+        if not self.token_path:
+            return True
+
+        if not self.role_arn:
+            error_msg = (
+                "The provided profile or the current environment is "
+                "configured to assume role with web identity but has no "
+                "role ARN configured. Ensure that the profile has the role_arn"
+                "configuration set or the AWS_ROLE_ARN env var is set."
+            )
+            raise Exception(error_msg)
+
+        return False
+
+    def _take_credentials_from_response(self, response: bytes) -> Dict[str, str]:
+        parsed_response: Dict[str, str] = {}
+
+        response_tree = ET.fromstring(response)
+        ns = {"sts": f"https://sts.amazonaws.com/doc/{AWS_STS_API_VERSION}/"}
+        credentials = response_tree.find(
+            "sts:AssumeRoleWithWebIdentityResult", ns
+        ).find("sts:Credentials", ns)
+
+        for cred in credentials:
+            _, _, tag = cred.tag.rpartition("}")
+            parsed_response[tag] = cred.text
+
+        return parsed_response
+
+    def _build_url(self) -> str:
+        params: Dict[str, Any] = {
+            "WebIdentityToken": self.identity_token,
+            "RoleArn": self.role_arn,
+            "RoleSessionName": self.role_session_name,
+            "Action": self.ACTION,
+            "Version": AWS_STS_API_VERSION,
+        }
+
+        return str(self.base_url.with_query(params))
+
+    @property
+    def identity_token(self) -> str:
+        if self._identity_token:
+            return self._identity_token
+
+        if not self.token_path:
+            raise Disabled()
+
+        with open(self.token_path) as token_file:
+            self._identity_token = token_file.read()
+
+        return self._identity_token
 
 
 # https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint.html


mrkovalchuk:
For now, it's just a workaround. I need approval on a concept before I write tests, etc. I do manual tests, and it works nicely.

src/aiodynamo/credentials.py
```diff
@@ -289,6 +293,81 @@ def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
...
+        if not config["role"]:
+            error_msg = (
+                "The provided profile or the current environment is "
+                "configured to assume role with web identity but has no "
+                "role ARN configured. Ensure that the profile has the role_arn"
+                "configuration set or the AWS_ROLE_ARN env var is set."
+            )
+            raise Exception(error_msg)
+
+        config["role_session_name"] = os.environ.get("AWS_ROLE_SESSION_NAME")
```
ojii, line 367: as with other credential loaders, these `os.environ` values should be settable class attributes (they can default to the value from `os.environ`).

src/aiodynamo/credentials.py
```diff
@@ -289,6 +293,81 @@ def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
...
+            key=Key(
+                id=credentials["AccessKeyId"],
+                secret=credentials["SecretAccessKey"],
+                token=credentials.get("Token", None),
+            ),
+            expires=parse_amazon_timestamp(credentials["Expiration"]),
+        )
+
+    def is_disabled(self) -> bool:
+        return False
```
ojii, line 330: looking at the implementation of `_load_config`, this should probably check if a `role` (and any other potentially necessary bits) has been provided. Since this loader is added to the default credentials loader list, it's important that `is_disabled` is a bit conservative so that the default loader can skip this if it's not configured.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ python = "^3.7"
 httpx = {version = ">=0.15.0 <1.0.0", optional = true}
 aiohttp = {version = "^3.6.2", optional = true}
 yarl = "^1.4.2"
+xmltodict = '^0.13.0'
```
ojii, line 28: I'm not too sure about this dependency. xml is scary to touch and from looking at the implementation this doesn't use something like [defusedxml](https://github.com/tiran/defusedxml) which hardens xml parsers in Python. It looks like you're also only interested in a few values in the XML returned from AWS (and the XML itself looks rather trivial), so using defusedxml + one of the stdlib xml parsers should not be too difficult.

src/aiodynamo/credentials.py
```diff
@@ -289,6 +293,81 @@ def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
...
+
+    def _parse_aws_response(self, response: bytes) -> dict:
+        parsed_response = xmltodict.parse(response)
+        return parsed_response["AssumeRoleWithWebIdentityResponse"]["AssumeRoleWithWebIdentityResult"]
+
+    def _build_url(self, identity_token: str, role: str, role_session_name: Optional[str]) -> str:
+        params = {
+            "WebIdentityToken": identity_token,
+            "RoleArn": role,
+            "RoleSessionName": role_session_name or f"default-role-name-{time.time()}",
```
ojii, line 340: is that really a reasonable default? I feel like this would be better as an explicit value users have to set when using this credentials loader.

src/aiodynamo/credentials.py
```diff
@@ -289,6 +293,81 @@ def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
...
+    ACTION = "AssumeRoleWithWebIdentity"
+    VERSION = "2011-06-15"
+
+    timeout: Timeout = 2
+    max_attempts: int = 3
+
+    base_url: URL = URL("https://sts.amazonaws.com/")
+
+    async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
+        config: dict = self._load_config()
```
ojii, line 308: I'd prefer this not to be an un-typed dict (`dict` is a very broad type, but you know what keys/values are gonna be there anyway). Either use a `TypedDict`, `dataclass` or since there's always three values from what I can see, a `tuple[str, str, str]`

src/aiodynamo/credentials.py
```diff
@@ -289,6 +293,81 @@ def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
...
+                secret=credentials["SecretAccessKey"],
+                token=credentials.get("Token", None),
+            ),
+            expires=parse_amazon_timestamp(credentials["Expiration"]),
+        )
+
+    def is_disabled(self) -> bool:
+        return False
+
+    def _parse_aws_response(self, response: bytes) -> dict:
```
ojii, line 332: as above, please don't use `dict` as a type hint.

src/aiodynamo/credentials.py
```diff
@@ -289,6 +293,81 @@ def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
     return thing
 
 
+# https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html
+@dataclass
+class AssumeRoleWithWebIdentityCredentials(MetadataCredentials):
+    ACTION = "AssumeRoleWithWebIdentity"
+    VERSION = "2011-06-15"
```
ojii, line 300: these should be module-level constants (with a more descriptive name), since they're not supposed to be changed.

ojii (changes requested):
Thank you for this pull request, it looks pretty good so far (but please see the comments), but as mentioned in the ticket, this will need tests.

ojii:
Please make sure to set up [pre-commit](https://pre-commit.com/) as this will run various code quality/standardization checks when you make a commit. right now some of them (eg mypy) are failing on this branch.

pyproject.toml
mrkovalchuk: I'm not a big fan of xml, so I use a library to avoid contact with it. I make an implementation that takes credentials from aws response. Comment please, I don't have enough practice with it.

src/aiodynamo/credentials.py
mrkovalchuk: botocore to do the same: https://github.com/boto/botocore/blob/develop/botocore/credentials.py#L788
I decide to save this behavior for consistency between libraries. It can help to move from botocore a little bit.

ojii:
So, I've just spent some time trying to see if I can figure out k8s/eks and the result is a resounding NO. For me to accept this PR though, I want to see this running in an EKS cluster.

Could you (or someone else who is interested in this being added) please provide a script/sequence of commands to run that create an EKS cluster, run some code in that cluster that shows this code works (eg show the output of `AssumeRoleWithWebIdentityCredentials(...).get_key(...)`) and then cleans everything up on AWS. The script/sequence of commands will be run in an environment that has admin access to an aws account, so don't worry about that part.

dimaqq:
Label this `help-wanted` maybe? 🙄

MarkusSintonen:
Hi @ojii, @mrkovalchuk! We have been running following implementation successfully for a long time in EKS: https://github.com/HENNGE/aiodynamo/compare/master...MarkusSintonen:support-AssumeRoleWithWebIdentity?expand=1

It has a very good test coverage (I adapted our unit tests to be compatible here) with the necessary parametrization for the credentials. It uses the JSON-format of the STS-API instead of XML. Im happy to open this one as an alternative PR instead of current PR (which doesnt have test coverage, also its missing retrying of certain STS client errors which are documented as retryable, etc.)

Here are (secrets-omitted) output we get in the EKS cluster:

```
% kubectl -n <namespace> exec -it deployment/<deployment> -- bash
$ python3 -m asyncio
asyncio REPL 3.11.3 (main, May  3 2023, 08:21:46) [GCC 10.2.1 20210110] on linux
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> import httpx
>>> import json
>>> from aiodynamo.http.httpx import HTTPX
>>> from extensions.aiodynamo.credentials import AssumeRoleWithWebIdentityEnvironmentCredentials
>>> from aiodynamo.http.types import Request
>>>
>>> # Show get_key output
>>> http = HTTPX(httpx.AsyncClient())
>>> creds = AssumeRoleWithWebIdentityEnvironmentCredentials()
>>> res = await creds.get_key(http)
>>> res.id
'<SECRET_OMITTED>'
>>> res.secret
'<SECRET_OMITTED>'
>>> res.token
'<SECRET_OMITTED>'
>>>
>>> # Show raw request and response
>>> str(creds.url)
'https://sts.eu-central-1.amazonaws.com/?Action=AssumeRoleWithWebIdentity&RoleArn=<SECRET_OMITTED>&RoleSessionName=aiodynamo-session-1684762674&Version=2011-06-15&WebIdentityToken=<SECRET_OMITTED>'
>>> response = await http(Request(method="POST", url=str(creds.url), headers={"Accept": "application/json"}, body=None))
>>> response.status
200
>>> print(json.dumps(json.loads(response.body), indent=2))
{
  "AssumeRoleWithWebIdentityResponse": {
    "AssumeRoleWithWebIdentityResult": {
      "AssumedRoleUser": {
        "Arn": "<SECRET_OMITTED>",
        "AssumedRoleId": "<SECRET_OMITTED>"
      },
      "Audience": "sts.amazonaws.com",
      "Credentials": {
        "AccessKeyId": "<SECRET_OMITTED>",
        "Expiration": 1684766752.0,
        "SecretAccessKey": "<SECRET_OMITTED>",
        "SessionToken": "<SECRET_OMITTED>"
      },
      "PackedPolicySize": null,
      "Provider": "<SECRET_OMITTED>",
      "SourceIdentity": null,
      "SubjectFromWebIdentityToken": "<SECRET_OMITTED>"
    },
    "ResponseMetadata": {
      "RequestId": "<UUID>"
    }
  }
}

```

mrkovalchuk:
Hi guys. I'm so sorry, and I want to be honest, I have no interest in that feature anymore. I will close this PR and if someone needs this implementation, fill free to continue. Maybe it will be me. Who knows. Sorry again.

ls-alexander-muehlbauer:
> [...] We have been running following implementation successfully for a long time in EKS: [`master...MarkusSintonen:support-AssumeRoleWithWebIdentity`?expand=1 (compare)](https://github.com/HENNGE/aiodynamo/compare/master...MarkusSintonen:support-AssumeRoleWithWebIdentity?expand=1)
> 
> [...] Im happy to open this one as an alternative PR instead of current PR (which doesnt have test coverage, also its missing retrying of certain STS client errors which are documented as retryable, etc.)

Hey @MarkusSintonen, would be actually great to see a PR for your implementation.
I think it's something this library can benefit from @ojii.

Let me know if I can be of any help.

ojii:
> Hey @MarkusSintonen, would be actually great to see a PR for your implementation. I think it's something this library can benefit from @ojii.
> 
> Let me know if I can be of any help.

My primary issue with all the proposed implementations of this is still that I cannot figure out how to test this in an AWS environment since I cannot figure out EKS. If an implementation of this is to be added to this repo, I want to see clear instruction on how to test this in the actual environment, which are doable by someone who has no understanding of k8s at all.

Alternatively, someone could provide this functionality as a third party package. The credentials system in aiodynamo is intentionally built in a way where you can provide your own mechanism to get credentials.

Issue #139, opened 2022-10-06
Auth Error With No Default Credentials


BTripp1986:
I have access to several AWS accounts via SSO and no default profile for AWS. When I sign in to AWS via SSO the `.aws/credentials` file is updated with a fresh Access Key, Secret Key, and Security Token.  

I specify which credentials I will use with the `AWS_PROFILE` env var, which is checked as part of the credentials chain lookup. If using boto3 or the aws-cli this method works. With this method I can sign in via SSO to dev and prod accounts. Then if I `AWS_PROFILE=dev aws s3 ls`(or similar call in boto3) it will run the command against the dev account. Then if I `AWS_PROFILE=prod aws s3 ls` it will run against prod.  

This doesn't seem to work with aiodynamo. In order to get it to work I need to create a default profile and then copy/paste the credentials from the desired profile into the default profile. Aiodynamo should respect the `AWS_PROFILE` env var.

ojii:
You can select which profile to use by using `aiodynamo.credentials.FileCredentials(profile=...)`, but I agree that it should default to load from `AWS_PROFILE`, so I'd accept a PR to that effect.

PR #146, opened 2022-12-18
add AWS_PROFILE env var to credentials chain

BTripp1986:
Adds support for using AWS_PROFILE env var in the credentials chain.  
This behavior is more in line with the behavior expected by users coming from boto3 or aiobotocore.
Added a unit test to cover this feature.

Relates to: Issue #139

### DIFF
--- a/src/aiodynamo/credentials.py
+++ b/src/aiodynamo/credentials.py
@@ -9,7 +9,7 @@ import os
 from dataclasses import dataclass, field
 from enum import Enum, auto
 from pathlib import Path
-from typing import Callable, Optional, Sequence, TypeVar
+from typing import Callable, Optional, Sequence, TypeVar, cast
 
 from yarl import URL
 
@@ -158,8 +158,10 @@ class FileCredentials(Credentials):
         self,
         *,
         path: Optional[Path] = None,
-        profile_name: str = "default",
+        profile_name: Optional[str] = None,
     ) -> None:
+        if profile_name is None:
+            profile_name = cast(str, os.environ.get("AWS_PROFILE", "default"))
         if path is None:
             path = Path.home().joinpath(".aws", "credentials")
         self.key = None
--- a/tests/unit/test_credentials.py
+++ b/tests/unit/test_credentials.py
@@ -180,7 +180,9 @@ async def test_disabled(monkeypatch: MonkeyPatch) -> None:
     assert not creds.is_disabled()
 
 
-async def test_file_credentials(fs: FakeFilesystem, http: HttpImplementation) -> None:
+async def test_file_credentials(
+    fs: FakeFilesystem, http: HttpImplementation, monkeypatch: MonkeyPatch
+) -> None:
     assert FileCredentials().is_disabled()
     fs.create_file(
         Path.home().joinpath(".aws", "credentials"),
@@ -201,6 +203,11 @@ async def test_file_credentials(fs: FakeFilesystem, http: HttpImplementation) ->
     credentials = FileCredentials(profile_name="my-profile")
     assert not credentials.is_disabled()
     assert await credentials.get_key(http) == Key(id="baz", secret="hoge")
+    monkeypatch.setenv("AWS_PROFILE", "my-profile")
+    credentials = FileCredentials()
+    assert not credentials.is_disabled()
+    assert await credentials.get_key(http) == Key(id="baz", secret="hoge")
+    monkeypatch.delenv("AWS_PROFILE", raising=False)
     custom_path = Path("/custom/credentials/file")
     assert FileCredentials(path=custom_path).is_disabled()
     fs.create_file(


src/aiodynamo/credentials.py
```diff
@@ -158,8 +158,9 @@ def __init__(
         self,
         *,
         path: Optional[Path] = None,
-        profile_name: str = "default",
+        profile_name: str = None,
     ) -> None:
+        profile_name = profile_name or os.environ.get("AWS_PROFILE", "default")
```
dimaqq, line 163: I wonder if I understand this change correctly:
When credentials are in a file, the profile selector is an env var?
And when credentials are from env, there's no selector?

BTripp1986: That is correct. This is in line with the behavior of other AWS libraries such as boto3 or the AWS-CLI. The credential chain still respects the load order, and credentials set with AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, etc. or a profile passed explicitly when initializing the client will take precedence over the AWS_PROFILE env var.  

The behavior this mimics is described in these AWS docs:  
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html

dimaqq:
cc @ojii the expert

ojii:
@BTripp1986 thank you for this pull request. the change looks good, but doesn't pass `mypy` checks. If you run `poetry run mypy` on your machine the error should be pointed out and trivially fixable.

It looks like we've had GitHub Actions misconfigured forever in this repo which is why checks don't run on external pull requests. Could you please rebase/merge your branch with the latest change on master? That should hopefully run the CI checks on this PR.

BTripp1986:
@ojii Sure thing, I should be able to take care of that at some point today or tomorrow as a worst case.  

re: Github Actions - I'm a DevOps Engineer so I'm pretty fluent in automation. I've never touched Github actions(I'm a Gitlab user) but I would like to get familiar with them and this seems like a good place to start. If someone could create an issue giving an overview of what's broken I can take a crack at it this weekend most likely.  

Not sure if #105 is related. I saw a closed MR(#106 ) related to that issue and I don't want to step on anyone's toes. Let me know if you would like my contribution.

ojii:
> If someone could create an issue giving an overview of what's broken I can take a crack at it this weekend most likely.

there _shouldn't_ be anything broken at the moment. when I did the transition from CircleCI to Github Actions I just misconfigured it to only run CI on pushes to this repo, but not external pull requests (which are pushes to another repo). 6f4011091827f53e9302c28b8b8ad16f128f507d should resolve that. 

> Not sure if #105 is related. I saw a closed MR(#106 ) related to that issue and I don't want to step on anyone's toes. Let me know if you would like my contribution.

that's more about automating releases which would be nice, #106 jumped the gun a bit there. If you want to try to get release automation to work with Github Actions, be my guest.

BTripp1986:
MyPy error fixed and branch has been rebased

src/aiodynamo/credentials.py
```diff
@@ -158,8 +158,9 @@ def __init__(
         self,
         *,
         path: Optional[Path] = None,
-        profile_name: str = "default",
+        profile_name: Optional[str] = None,
     ) -> None:
+        profile_name = profile_name or os.environ.get("AWS_PROFILE", "default")
```
ojii, line 163: if you assign to a different variable name here in replace the `or` with an `if`, you don't need the `str(...)` in line 178.

ojii: actually, this seems like a mypy [bug](https://github.com/python/mypy/issues/14338) so maybe just using `typing.cast` here and removing the `str(...)` below would be fine too.

dimaqq: For a quick fix:

`pro = pro if pro else od.env(...)`

or possibly more legible / correct:

```py
if profile_name is None:
    profile_name = os.env(...)
```

src/aiodynamo/credentials.py
```diff
@@ -174,7 +175,7 @@ def __init__(
             return
 
         try:
-            profile = parser[profile_name]
+            profile = parser[str(profile_name)]
```
dimaqq, line 178: is this str cast only for mypy, or is there some other reason behind it?

src/aiodynamo/credentials.py
ojii: you cannot narrow an existing name in mypy... at least not that way https://mypy-play.net/?mypy=latest&python=3.11&gist=774404f9e0075cf360b4b7e54b013546

dimaqq: Right, good point. I should have remembered that.
`x if x else y` it is then.

ojii: how is mypy okay with that but not `or` :( 

anyway, inlined looks good to me if it pacifies mypy

src/aiodynamo/credentials.py
BTripp1986: Simply for mypy

src/aiodynamo/credentials.py
BTripp1986: Added `typing.cast` to take care of that. I initially wrote in as an inline `x if x is not None else y` but when formatting was applied it looked hideous so I went with a multi-line `if` statement.

ojii:
@BTripp1986 thanks for the mypy fixes, looks like there's just one small isort issue left to fix. I recommend you set up [pre-commit](https://pre-commit.com/) locally so that the static checks run before you commit. Once that's resolved I'll merge this and make a release.

BTripp1986:
isort issue has been fixed. Thanks for walking me through this.

ojii:
Thanks again @BTripp1986, this has been released as 22.12

PR #149, opened 2023-02-07
Add debug helper to Condition and UpdateExpression

ojii:
These can be useful when trying to figure out why a complex expression isn't doing what you think it should. 

Also added a pre-commit hook for the unit tests since those run super quick anyway. and tried to make the docs a tiny bit nicer too lok at (at some point probably need to change the theme or something, quite a few of these pages look very crowded)

### DIFF
--- a/src/aiodynamo/expressions.py
+++ b/src/aiodynamo/expressions.py
@@ -4,11 +4,22 @@ import abc
 import decimal
 from dataclasses import dataclass, field
 from itertools import chain, count
-from typing import Any, Dict, Iterator, List, Optional, Sequence, Set, Tuple, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Sequence,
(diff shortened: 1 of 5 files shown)


rowr111 (approved):
thank you!!!

src/aiodynamo/expressions.py
```diff
@@ -378,6 +389,14 @@ def __invert__(self) -> Condition:
     def encode(self, params: Parameters) -> str:
         pass
 
+    def debug(self, numeric_type: Callable[[str], Any] = float) -> str:
```
dimaqq, line 392: How about using this in `__repr__`?

dimaqq (approved):

src/aiodynamo/expressions.py
ojii: I actually quite like the current dataclass-provided `__repr__`. I often find it easier to read when there's lots of &'ing and |'ing happening.

PR #151, opened 2023-02-20
Add ScyllaDB Alternator to test suite

ojii:
Adds ScyllaDB Alternator to our suite of implementations to test with.

Changed how the test suite finds local implementations and made it easier to run tests with multiple implementations at once (`DYNAMODB_URL` -> `DYNAMODB_URLS`), though `DYNAMODB_URL` will still work (since it's useful for GHA). 

Scylla handles `DescribeTable` a bit different from other implementations, see https://github.com/scylladb/scylladb/issues/12930, but that's not too bad since it found a bug in `describe_table` (returning `throughput=None` instead of `throughput=PayPerRequest()`), so I've updated the test accordingly, though that requires setting the "flavor" to "scylla" for those tests (either via `DYNAMODB_FLAVOR=scylla` or `DYNAMODB_URLS='scylla=http://endpoint,scylla'`

Also used this chance to add 3.11 testing.


ojii:
Predicatbly, I broke the GHA workflow file so badly it doesn't even run any tests.

dimaqq:
quite a few changes are mixed in together... from newer python to scylla to code fixes

i guess that's why it's a draft :)

ojii:
> quite a few changes are mixed in together... from newer python to scylla to code fixes
> 
> i guess that's why it's a draft :)

It's a draft because I broke Github Actions somehow and Github decides if you break it, that's not "failing tests" but "no tests, everything ok!" 🤦

dimaqq:
`The workflow is not valid. .github/workflows/workflow.yml (Line: 115, Col: 9): Unexpected value 'command'`

Perhaps the value can only be a string and not a list?

GHA are opt-in. Set up branch protection rules for the PR target branch and require specific tests to pass to avoid silent failures.

FurqanHabibi:
I wouldn't recommend using GHA service container and this is one of the reason, it does not yet support passing arguments to the container 🤯 

Instead just use docker-compose and run it in a step, that's what we do in [unreadchecker](https://github.com/HENNGE/shobosso-unreadchecker/blob/dev/.github/workflows/workflow.yml#L66) for example.

FurqanHabibi:
And about the GHA not running any check, yes because the workflow yaml is invalid, so it is not regarded as a check.

FurqanHabibi:
I've setup the docker-composes for the tests in [this branch ](https://github.com/HENNGE/aiodynamo/tree/support-scylla-compose). I can merge it to this branch if you want. Do not that the tests are not passing 100% yet.

ojii:
> I've setup the docker-composes for the tests in [this branch ](https://github.com/HENNGE/aiodynamo/tree/support-scylla-compose). I can merge it to this branch if you want. Do note that the tests are not passing 100% yet.

Thank you 💯

.github/workflows/workflow.yml
```diff
@@ -2,149 +2,128 @@ name: workflow
...
           key: ${{ github.sha }}-${{ matrix.python-version }}
-  pytest-dynalite:
+
+  pytest:
     needs: lock
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        implementation: ["dynalite", "dynamodb-local", "localstack"]
```
FurqanHabibi, line 32: You can set the matrix element to be a dictionary (or object as it's called in GHA) like so:
```yml
implementation: [{flavor: "other", name: "dynalite"}, {flavor: "other", name: "dynamodb-local"}, {flavor: "other", name: "localstack"}, {flavor: "scylla", name: "scylla"}]
```

Now you can refer them in env like so:
```yml
env:
  ...
  DYNAMODB_FLAVOR: ${{ matrix.implementation.flavor }}
  DYNAMODB_NAME: ${{ matrix.implementation.name }}
```

So now we can get rid of `pytest-scylla`.

ojii: huh their documentation could really make this clearer...

FurqanHabibi: Couldn't agree more

.github/workflows/workflow.yml
```diff
@@ -2,149 +2,102 @@ name: workflow
...
       DYNAMODB_URL: http://localhost:8000
+      DYNAMODB_FLAVOR: ${{ matrix.implementation.flavor }}
+      DYNAMODB_NAME: ${{ matrix.implementation.name }}
       AWS_ACCESS_KEY_ID: dummy
       AWS_SECRET_ACCESS_KEY: dummy
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - run: docker compose -f .github/docker-compose/docker-compose.${{ matrix.implementation.name }}.yml up -d
```
FurqanHabibi, line 42: Since you already set an env for it, you can just use that instead:
```yml
- run: docker compose -f .github/docker-compose/docker-compose.${DYNAMODB_NAME}.yml up -d
```

ojii: tbh i prefer the current way it's written.

FurqanHabibi: But currently we need to put `${{ matrix.implementation.name }}` two times right?

ojii: yea but it makes it clear to me where this value is coming from

.github/workflows/workflow.yml
```diff
@@ -2,149 +2,102 @@ name: workflow
...
+        implementation: [{flavor: "other", name: "dynalite"}, {flavor: "other", name: "dynamodb-local"}, {flavor: "other", name: "localstack"}, {flavor: "scylla", name: "scylla"}]
     runs-on: "ubuntu-latest"
-    services:
-      dynamodb:
-        image: amazon/dynamodb-local:latest
-        ports:
-          - 8000:8000
     env:
       DYNAMODB_URL: http://localhost:8000
+      DYNAMODB_FLAVOR: ${{ matrix.implementation.flavor }}
```
FurqanHabibi, line 36: After a bit of playing around, there is in fact a way to do ternary operation, like so:
```yml
DYNAMODB_FLAVOR: ${{ matrix.implementation.flavor != NaN && matrix.implementation.flavor || 'other' }}
```
Take note that string in a GHA expression _must_ use single quote.

.github/workflows/workflow.yml
```diff
@@ -2,149 +2,102 @@ name: workflow
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lock:
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
```
FurqanHabibi, line 13: I'm not too fond of this solution, but there is a way to set this once using repository variables. You set it [in the github UI](https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository), similar to setting repository secrets. Then you can access it from anywhere in your workflow file like so:
```yml
python-version: ${{ vars.<YOUR_VAR_NAME> }}
```

ojii: 🤔 so that way we could add/remove versions without changing the workflow file, i can see how this could be useful. But I think I prefer the file to be as self-explanatory as possible

FurqanHabibi: Yes exactly. Pros and cons.

FurqanHabibi:
@ojii This should be good to merge right?

ojii:
> @ojii This should be good to merge right?

yea if we can fix ci

dimaqq:
Woo-hoo!!!

PR #155, opened 2023-06-02
Set timeout-minutes of 10 min to all jobs

FurqanHabibi:
To prevent broken jobs from running indefinitely.


FurqanHabibi:
I think we should merge this first, to prevent others from wasting gha minutes, since this is an OSS.

Later we can consult jonas on how to actually fix the poetry issue 😄

hanscg:
GHA is free for public repo so we're just wasting Github's money 😬

PR #156, opened 2023-06-28
Fix poetry lock running forever

FurqanHabibi:
Dependency resolution took a long time because poetry chose the wrong version of urllib3.
Specifying the urllib3 version solve the resolution.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ typing_extensions = { version = "^3.7", python = "< 3.8" }
 httpx = ["httpx"]
 aiohttp = ["aiohttp"]
 
-[tool.poetry.dev-dependencies]
+[tool.poetry.group.dev.dependencies]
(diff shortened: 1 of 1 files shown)


dimaqq (approved):
:OMG: this is hilarious 🙃

stupoid:
@hanscg 
Just checking, is it ok to merge this in?

hanscg (approved):

hanscg:
@stupoid 
It looks good to me, but I don't exactly manage this repo 🙇 

@ojii Any comments? Otherwise I'll merge this

ojii:
> @ojii Any comments? Otherwise I'll merge this

I'm a little unhappy about setting an upper bound on urllib3 instead of a lower one but if this is needed, i don't have an objection to merging it.

hanscg:
Let's merge and see if we can do anything about urllib later 🙏

PR #158, opened 2023-10-02
Properly support creating pay-per-request tables

alexjurkiewicz:
When creating a table in pay-per-request billing mode, don't specify provisioned throughput for GSIs.

It might be better to create two classes of the Table type, one for each billing mode. But that's a major (probably breaking) refactor, and the AWS API doesn't do this either.

### DIFF
--- a/src/aiodynamo/models.py
+++ b/src/aiodynamo/models.py
@@ -8,7 +8,17 @@ import time
 from dataclasses import dataclass
 from enum import Enum, unique
 from itertools import count
-from typing import Any, AsyncIterable, Dict, Iterable, Iterator, List, Optional, cast
+from typing import (
+    Any,
+    AsyncIterable,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    cast,
+    Union,
+)
 
 from .expressions import Parameters, ProjectionExpression
 from .types import (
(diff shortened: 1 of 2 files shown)


src/aiodynamo/models.py
```diff
@@ -129,13 +129,13 @@ def encode(self) -> EncodedLocalSecondaryIndex:
 
 @dataclass(frozen=True)
 class GlobalSecondaryIndex(LocalSecondaryIndex):
-    throughput: Throughput
+    throughput: Optional[Throughput]
 
     def encode(self) -> EncodedGlobalSecondaryIndex:
         # Cast due to https://github.com/python/mypy/issues/4122
         return cast(
             EncodedGlobalSecondaryIndex,
```
ojii, line 137: please also update the type definition of `EncodedGlobalSecondaryIndex`, it should now be [`total=False`](https://docs.python.org/3/library/typing.html#typing.TypedDict) since `ProvisionedThroughput` is not always set.

ojii:
For reference: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_CreateTable.html

Oddly enough, the docs don't explain this clearly, it just mentions that `ProvisionedThroughput` on a GSI spec is optional...

Anyway, thank you for noticing this and the PR.

src/aiodynamo/models.py
alexjurkiewicz: done, thanks!

alexjurkiewicz:
Done

src/aiodynamo/models.py
```diff
@@ -129,13 +129,13 @@ def encode(self) -> EncodedLocalSecondaryIndex:
...
-            {**super().encode(), **self.throughput.encode()},
+            {**super().encode(), **self.throughput.encode() if self.throughput else {}},
```
ojii, line 138:
```suggestion
            {**super().encode(), **(self.throughput.encode() if self.throughput else {})},
```

ojii:
mypy is being silly, i'll fix it.

alexjurkiewicz:
I was looking too 😊 My fix is a little ugly, but satisfies the checker

ojii:
merged locally, so closing this PR. Thank you.

alexjurkiewicz:
Thanks for the help!

Issue #157, opened 2023-08-31
InstanceMetadataCredentials does not work with IMDSv2


y26805:
InstanceMetadataCredentials attempts to get metadata about the EC2 instance by a GET request to the AWS endpoint `"http://169.254.169.254"`
https://github.com/HENNGE/aiodynamo/blob/62e6e5a77604d4cc98241678c55e338659c3cd81/src/aiodynamo/credentials.py#L357C7-L388

However, this does not work with the newer way of how AWS does this ("IMDSv2" = Instance Metadata Service Version 2). 

from [official docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html)
> 
> For IMDSv2, you must use /latest/api/token when retrieving the token.
> 
> Issuing PUT requests to any version-specific path, for example /2021-03-23/api/token, will result in the metadata service returning 403 Forbidden errors. This behavior is intended.
> 
> If IMDSv2 is required, IMDSv1 does not work.
> 
> You can check if IMDSv2 is required for an instance, as follows: Select the instance to view its details, and check the value for IMDSv2. The value is either Required (only IMDSv2 can be used) or Optional (IMDSv2 and IMDSv1 can be used).
> 

If the EC2 instance uses the [Amazon Linux 2023](https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2.html) Image, IMDSv2 is required by default.

> By default, any instances launched with the AL2023 AMI will require the use of IMDSv2-only

(IMDSv2 can be turned off in a very hidden options menu... for now)

would be great if aiodynamo can also support credentials retrieval via IMDSv2

y26805:
looks like we can just switch to the IMDSv2 method, since it's either
- IMDSv2 only; or
- support both IMDSv1 and v2

> Both IMDSv1 and IMDSv2 will be available and enabled by default, and customers can choose which they will use. The IMDS can now be restricted to v2 only, or IMDS (v1 and v2) can also be disabled entirely. AWS recommends adopting v2 and restricting access to v2 only for added security. 

(from https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/

ojii:
> * IMDSv2 only; or
> * support both IMDSv1 and v2

that's for new instances presumably, I'd expect that depending on how/when/... an instance got created/launched, it might only have v1?

[...]

PR #159, opened 2023-10-02
Support IMDSv2 to retrieve metadata

y26805:
Fixes https://github.com/HENNGE/aiodynamo/issues/157

Tested on an IMDSv2-only instance on AWS EC2, and it works 🎉


ojii:
shouldn't we support both flavors of IMDS?

y26805:
> shouldn't we support both flavors of IMDS?

IMDSv2 is supposed to be more secure (so AWS will probably remove v1 support at some point). Also, it's currently not possible to only allow IMDSv2, based on AWS documentation.
https://github.com/HENNGE/aiodynamo/issues/157#issuecomment-1742510749

src/aiodynamo/credentials.py
```diff
@@ -407,6 +408,32 @@ async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
...
+                max_attempts=self.max_attempts,
+                timeout=self.timeout,
+                request=Request(
+                    method="PUT",
+                    url=str(token_url),
+                    headers={
+                        "X-aws-ec2-metadata-token-ttl-seconds": str(
+                            session_duration_seconds
+                        )
+                    },
```
ojii, line 427: shouldn't we cache this token if it has a TTL?

src/aiodynamo/credentials.py
```diff
@@ -354,9 +355,94 @@ async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
 
 # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
 @dataclass
-class InstanceMetadataCredentials(MetadataCredentials):
+class InstanceMetadataCredentialsWithImdsV2(MetadataCredentials):
```
ojii, line 358: suuuper nitpicky, but "InstanceMetadata" and "Imds" are redundant. I think `InstanceMetadataCredentialsV1` and `InstanceMetadataCredentialsV2` would be more appropriate

tests/unit/test_credentials.py
```diff
@@ -129,15 +152,23 @@ async def test_ec2_instance_metdata_credentials(
...
-    http: HttpImplementation, instance_metadata_server: InstanceMetadataServer
+    http: HttpImplementation,
+    instance_metadata_server: InstanceMetadataServer,
+    use_v2: bool,
 ) -> None:
+    model = (
+        InstanceMetadataCredentialsWithImdsV2
+        if use_v2
+        else InstanceMetadataCredentialsWithImdsV1
+    )
```
ojii, line 165: why not parametrize model using `parametrize("model", [InstanceMetadataCredentialsWithImdsV2, InstanceMetadataCredentialsWithImdsV1])`?

tests/unit/test_credentials.py
```diff
@@ -60,7 +65,9 @@ async def credentials_handler(self, request: web.Request) -> web.Response:
 
 
 @pytest.fixture
-async def instance_metadata_server() -> AsyncGenerator[InstanceMetadataServer, None]:
+async def instance_metadata_server(
+    use_v2: bool = True,
```
ojii, line 69: that's not how fixtures work... either you need to use marks, another fixture, or just always include v2

src/aiodynamo/credentials.py
y26805: yeah i was thinking how to name them.. `InstanceMetadataCredentialsV2`/V1 sounds good. will change

tests/unit/test_credentials.py
y26805: i was thinking about having 2 fixtures, but it feels like an overkill... so yeah let me revert that

ojii (approved):

PR #154, opened 2023-06-02
Update maximum operations per transaction to 100

stupoid:
This resolves https://github.com/HENNGE/aiodynamo/issues/153 by updating the max values from 25 to 100.

PR #160, opened 2023-10-02
Update maximum operations per transaction to 100

ojii:
duplicate of #154 to try to get CI checks to run


ojii:
okay, it fails because poetry dropped 3.7 support (which we should too). let me open a different PR to update some stuff and then rebase this one later

PR #163, opened 2023-10-02
Add GHA for PyPi release

FurqanHabibi:
As with AAPNS, this will automatically publish a new pypi release when a Github release is created.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,58 @@
+name: aiodynamo release workflow
+
+on:
+  release:
+    types: [released] # only triggered on release, not pre-release or draft release
+
+permissions:
+  contents: read
+  id-token: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  PYTHON_VERSION_MIN: 3.8
+  TARGET_BRANCH: ${{ github.event.release.target_commitish }}
+  DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+  TAG_NAME: ${{ github.ref_name }}
+
+jobs:
+  release:
+    name: Release to PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check whether release target branch is default branch
+        if: env.TARGET_BRANCH != env.DEFAULT_BRANCH
+        uses: actions/github-script@v6
+        with:
+          script: |
+            core.setFailed('Target branch (${{ env.TARGET_BRANCH }}) is not default branch (${{ env.DEFAULT_BRANCH }}).')
+      
+      - uses: actions/checkout@v3
+      
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION_MIN }}
+          
+      - run: curl -sSL https://install.python-poetry.org | python3 -
+      
+      - run: echo "PKG_VER=$(poetry version --short)" >> $GITHUB_ENV
+      
+      - name: Check whether current tag matches package version
+        if: env.TAG_NAME != env.PKG_VER
+        uses: actions/github-script@v6
+        with:
+          script: |
+            core.setFailed('Tag name (${{ env.TAG_NAME }}) does not match package version (${{ env.PKG_VER }}).')
+      
+      - name: Build package
+        run: poetry build
+        
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,7 @@ jobs:
       
       - uses: actions/setup-python@v4
         with:
-          python-version: ${{ env.PYTHON_VERSION_MIN }}
+          python-version: ${{ env.PYTHON_VERSION }}
           
       - run: curl -sSL https://install.python-poetry.org | python3 -
       
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ defaults:
     shell: bash
 
 env:
-  PYTHON_VERSION_MIN: 3.8
+  PYTHON_VERSION: 3.11
   TARGET_BRANCH: ${{ github.event.release.target_commitish }}
   DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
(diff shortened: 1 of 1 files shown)


FurqanHabibi:
@ojii Let's try this? Perhaps we can create a new release while at it?

ojii:
A couple more changes are coming soon so I want to wait for those

ojii:
@FurqanHabibi just to check, after merging, to make a release the procedure is:

1. Change version in pyproject.toml
2. Update docs/changelog.rst
3. make a commit and push
4. create a release on github 

correct?

FurqanHabibi:
100% correct 👍 

Make sure release tag == pyproject.toml version

dimaqq:
> 1. Change version in pyproject.toml
> 2. Update docs/changelog.rst
> 3. make a commit and push
> 4. create a release on github

I recommend `release-please` github action to automate all these steps.
It's pretty easy to set up and works pretty well.
(maybe expect one WAT moment, but still rather easy to figure out)

It does require following https://www.conventionalcommits.org/en/v1.0.0/ though.

dimaqq:
Edit: CalVer support is still on TODO list: https://github.com/googleapis/release-please/issues/1848

ojii:
> Edit: CalVer support is still on TODO list: [googleapis/release-please#1848](https://github.com/googleapis/release-please/issues/1848)

that's a no from me then.

dimaqq:
https://github.com/python-semantic-release/python-semantic-release then?

I haven't tried it, but I think `version_pattern` can be crafted smth like `YY.patch`?

ojii (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
+name: aiodynamo release workflow
+
+on:
+  release:
+    types: [released] # only triggered on release, not pre-release or draft release
+
+permissions:
+  contents: read
+  id-token: write
```
dimaqq, line 9: That's unusual, what does it really do / why is it needed?

FurqanHabibi: It's required for PyPI trusted publishing
https://github.com/pypa/gh-action-pypi-publish#trusted-publishing

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+          script: |
+            core.setFailed('Target branch (${{ env.TARGET_BRANCH }}) is not default branch (${{ env.DEFAULT_BRANCH }}).')
+      
+      - uses: actions/checkout@v3
+      
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION_MIN }}
+          
+      - run: curl -sSL https://install.python-poetry.org | python3 -
```
dimaqq, line 43: Minor:
There's a github action available for that, which allows pinning poetry version.

FurqanHabibi: We can also pin the version here with
```
curl -sSL https://install.python-poetry.org | POETRY_VERSION=X.X.X python3 -
```

I just prefer to use the official installation command rather than depend on some third party action, especially since it's only a one liner.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/github-script@v6
+        with:
+          script: |
+            core.setFailed('Target branch (${{ env.TARGET_BRANCH }}) is not default branch (${{ env.DEFAULT_BRANCH }}).')
+      
+      - uses: actions/checkout@v3
+      
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION_MIN }}
```
dimaqq, line 41: I'm curious about the rationale for using older Python.
Is it intended to ensure that the built artefacts actually work with older Python when they are downloaded from PYPI?
In that case, running tests would be better.

Otherwise, all this does is sets up what Python is used to run poetry.
So, why not 3.11?

.github/workflows/release.yml
FurqanHabibi: And about the version, I just thought using the latest poetry version should be fine for this.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  PYTHON_VERSION_MIN: 3.8
+  TARGET_BRANCH: ${{ github.event.release.target_commitish }}
+  DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
```
dimaqq, line 22: Super-minor and debatable: since github-script is used anyway, why not evaluate the event bits directly in the script block?

.github/workflows/release.yml
FurqanHabibi: Since poetry is used for building, I just thought using the lowest possible python is better?
As set in the pyproject.toml
https://github.com/HENNGE/aiodynamo/blob/master/pyproject.toml#L24

Although I'm not that too familiar with how poetry performs the build process, so any input is appreciated.

ojii: we might actually run into issues with this since we don't pin poetry and poetry drops support for older python versions once they get EOL'ed, so in about a year this will likely break. Using the highest version here might be better.

.github/workflows/release.yml
FurqanHabibi: The script block is there really just to early-exit the workflow, since it's the easiest way to do it at the moment
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#:~:text=echo-,core.setFailed,-Used%20as%20a

Otherwise, I think it's better to keep all the logic in GitHub Action yaml and not in github-script.

.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/github-script@v6
+        with:
+          script: |
+            core.setFailed('Target branch (${{ env.TARGET_BRANCH }}) is not default branch (${{ env.DEFAULT_BRANCH }}).')
+      
+      - uses: actions/checkout@v3
+      
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION_MIN }}
```
dimaqq, line 41: You're renamed the env var declaration...
```suggestion
          python-version: ${{ env.PYTHON_VERSION }}
```

dimaqq (approved):

ojii:
thanks, i'll do a release and try this thing then

ojii:
🎉 https://pypi.org/project/aiodynamo/23.10/

PR #165, opened 2023-10-03
Better token refresh

ojii:
Extract token refresher from MetadataCredentials and re-use it in imdsv2

### DIFF
--- a/src/aiodynamo/credentials.py
+++ b/src/aiodynamo/credentials.py
@@ -9,7 +9,16 @@ import os
 from dataclasses import dataclass, field
 from enum import Enum, auto
 from pathlib import Path
-from typing import Callable, Optional, Sequence, TypeVar, cast
+from typing import (
+    Awaitable,
+    Callable,
+    Generic,
+    Optional,
+    Sequence,
+    TypeVar,
+    Union,
+    cast,
+)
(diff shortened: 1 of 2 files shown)


src/aiodynamo/credentials.py
```diff
@@ -370,17 +419,21 @@ class InstanceMetadataCredentialsV2(MetadataCredentials):
         ).lower()
         == "true"
     )
-    auth_token: str = ""
+    token_session_duration_seconds: int = 5 * 60
```
y26805, line 422: actually, should we set this to max which is 6 hours (6 * 60 * 60)? since now the refresh threshold is like 15 min, it will always be refreshing if it were 5 min

ojii: that sounds good. should we also check that it's <=6h somewhere to avoid mistakes?

y26805 (approved):
LGTM!! thank you

y26805:
(will also test this on AWS Linux 2023 in my sandbox.. will report back

y26805:
> (will also test this on AWS Linux 2023 in my sandbox.. will report back

Tested on an IMDSv2-only EC2 instance, works great 🎉

PR #168, opened 2023-10-19
Fix timing bug in Refreshable, optimize ChainCredentials

ojii:
Refreshable had a timing bug where refreshing the token could result in an invalid state if invalidate() was called at the same time. This is due to the Event Loop being able to schedule other coroutines to run between the current value being set in _refresh() and the value being checked in get(). By moving the assignment from _refresh() to get(), no other coroutines can run between the setting and checking, preventing the class from ending up in an invalid state.

ChainCredentials would needlessly check many candidates every time, even if it had already done so previously. Since running code cannot move to different environments, this check is unneccessary and inefficient, so instead ChainCredentials now remembers when a provider has succeeded and will only try that one again.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,5 +33,5 @@ repos:
         language: system
         types: [python]
         pass_filenames: false
-        args: ['check', 'src', 'tests']
+        args: ['check', '--fix', 'src', 'tests']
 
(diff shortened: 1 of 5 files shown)


y26805 (approved):
tested on a AWS EC2 instance (IMDSv2), loader and credentials refresh works well

PR #171, opened 2023-12-14
Implement consistent handling of TransactionCanceled errors.

### LINKED ISSUE #170: Awkward API for handling `TransactionCanceled` from `transact_write_items`
# Overview

When using `transact_write_items` with multiple items, handling errors is inconsistent.

If the first item in the list caused the transaction to abort and it is one of the handled
error types in aiodynamo:

https://github.com/HENNGE/aiodynamo/blob/922eb16fefcf94801897ccfe6c602fb063afa3c3/src/aiodynamo/errors.py#L172-L200

the cause is extracted and raised back up to the user:

https://github.com/HENNGE/aiodynamo/blob/922eb16fefcf94801897ccfe6c602fb063afa3c3/src/aiodynamo/errors.py#L211-L212

https://github.com/HENNGE/aiodynamo/blob/922eb16fefcf94801897ccfe6c602fb063afa3c3/src/aiodynamo/errors.py#L218-L223

If it is any other item in the list, it is unhandled and the `TransactionCanceled` is bubbled up.

This looks like it stems from this comment:

https://github.com/HENNGE/aiodynamo/pull/132#discussion_r939773600

This requires code using this API to handle errors in two different ways - if the code
cares about the cause of the cancellation.

Say you want to update 3 items - each has a conditional check and but you want to handle
the check failing differently for the three items.

To do this, you must have an `except` block for `ConditionalCheckFailed` - that handles the
first item in the list, then you must also have an `except` block for `TransactionCanceled`
and inspect the `CancellationReasons` - if the second or third reason has code
`ConditionalCheckFailed`, then that item was the condition check which aborted the transaction.

For my use case, it would be clearer and simpler if we always raise the `TransactionCanceled`
error. The other option, extending the inspection to all items in the `CancellationReasons`,
could work, but has the disadvantage that there is not enough information to correlate back
to the item which caused the rollback. For example, a rollback caused by the 5th item looks
like this:

```
{'__type': 'com.amazonaws.dynamodb.v20120810#TransactionCanceledException', 'CancellationReasons': [{'Code': 'None'}, {'Code': 'None'}, {'Code': 'None'}, {'Code': 'None'}, {'Code': 'ConditionalCheckFailed', 'Message': 'The conditional request failed'}], 'Message': 'Transaction cancelled, please refer cancellation reasons for specific reasons [None, None, None, None, ConditionalCheckFailed]'}
```

If the existing conversion extends to all, you could handle the `ConditionalCheckFailed` in an
except block, but would not know which item caused it.

Before I open a PR, I would like to ask, how do you handle this case in your code? Maybe you
haven't needed to handle multiple items in different ways?

Depending on how we change this, it is potentially a breaking API change. One option to avoid
that could be to make the client constructor take an option to configure the unwrapping behaviour.
It defaults to this legacy behaviour, but also supports always unwrap and never unwrap.

Thoughts?

aclemons:
Previously, the code handling errors would raise the cause of the
cancellation, for example ConditionalCheckFailed or ThrottlingError, if
it was caused by the first item in the list of items in the transaction.

If the failure was caused by any other item in the list, the caller
would see the TransactionCanceled.

This makes is awkward for calling code to handle errors when interested
in the exact item which caused the failure - it needs two `except`
blocks for handling essentially the same thing. One for when it is the
first item and another for any other item.

This change removes any special handling and bubbles the
`TransactionCanceled` error up to the caller. I've added a typed
value for the cancellation reasons. These are documented here:

https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html#API_TransactWriteItems_Errors

This should allow clients to introspect the errors, if needed, with a
typed object.

Closes #170


aclemons:
If we don't like the typed cancellation reasons, I can revert that back out and
leave that for library consumers to handle themselves.

src/aiodynamo/errors.py
```diff
@@ -1,5 +1,10 @@
 import json
-from typing import Any, Dict
+from typing import Any, Dict, List, Optional, TypedDict
+
+
+class CancellationReason(TypedDict):
+    Code: Optional[str]
+    Message: Optional[str]
```
ojii, line 7: to keep things consistent with how other parts of aiodynamo work, this should be a (frozen) dataclass using pythonic identifiers (no uppercase for attributes).

src/aiodynamo/errors.py
```diff
@@ -110,7 +115,12 @@ class PointInTimeRecoveryUnavailable(AIODynamoError):
 
 
 class TransactionCanceled(AIODynamoError):
-    pass
+    def __init__(self, body: Dict[str, Any]):
```
ojii, line 118: could you add a type hint on the class for `cancellation_reasons` to make it clear that this is part of the public api.

ojii (changes requested):
thank you. looks pretty good, just some minor stylistic changes I'd appreciate

src/aiodynamo/errors.py
aclemons: Should I also convert:


```
No Errors:
    Code: None
    Message: null
```

to just be a `None` value in the array or leave it as a `CancellationReason(code="None", message=None)`?

ref: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html#API_TransactWriteItems_Errors

ojii: using `None` for those cases seems like a good idea

aclemons:
Thanks, I had this on my TODO and let it slip out of focus in the new year. Thanks for tidying up and merging.

PR #173, opened 2024-01-04
Flatten AND and OR conditions

ojii:
Previously, AND and OR conditions with more than two elements would lead to nested ANDs and ORs, making the resulting expression much more complicated and harder to read/understand. This change flattens AndCondition and OrCondition objects when they're combined.

Context: had to debug some complicated conditions (~10 AND's) and the extreme nesting makes it super hard to debug/read them.

~Not a _huge_ fan of the name `MinLen2AppendOnlyList` so if someone has a better suggestion I'd love to hear it.~

### DIFF
--- a/src/aiodynamo/utils.py
+++ b/src/aiodynamo/utils.py
@@ -5,6 +5,7 @@ import datetime
 import decimal
 import logging
 from collections import abc as collections_abc
+from dataclasses import dataclass
 from functools import reduce
 from typing import (
     TYPE_CHECKING,
@@ -12,6 +13,9 @@ from typing import (
     Awaitable,
     Callable,
     Dict,
+    Generator,
+    Generic,
+    Iterable,
     List,
     Mapping,
     Set,
@@ -197,3 +201,43 @@ def deparametetrize(
     for key, value in params.names.items():
         expression = expression.replace(key, value)
     return expression
+
+
+@dataclass(frozen=True)
+class MinLen2AppendOnlyList(Generic[T]):
+    first: T
+    second: T
+    rest: tuple[T, ...]
+
+    @classmethod
+    def create(cls, first: T, second: T, *rest: T) -> MinLen2AppendOnlyList[T]:
+        return cls(first, second, rest)
+
+    def appending(self, value: T) -> MinLen2AppendOnlyList[T]:
+        return MinLen2AppendOnlyList(self.first, self.second, (*self.rest, value))
+
+    def extending(self, values: Iterable[T]) -> MinLen2AppendOnlyList[T]:
+        return MinLen2AppendOnlyList(self.first, self.second, (*self.rest, *values))
+
+    def __contains__(self, item: Any) -> bool:
+        return item == self.first or item == self.second or item in self.rest
+
+    def __getitem__(self, index: int) -> T:
+        if index == 0:
+            return self.first
+        elif index == 1:
+            return self.second
+        return self.rest[index - 2]
+
+    def __len__(self) -> int:
+        return len(self.rest) + 2
+
+    def __iter__(self) -> Generator[T, None, None]:
+        yield self.first
+        yield self.second
+        yield from self.rest
+
+    def __reversed__(self) -> Generator[T, None, None]:
+        yield from reversed(self.rest)
+        yield self.second
+        yield self.first
--- a/src/aiodynamo/utils.py
+++ b/src/aiodynamo/utils.py
@@ -5,7 +5,6 @@ import datetime
 import decimal
 import logging
 from collections import abc as collections_abc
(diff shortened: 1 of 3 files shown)


src/aiodynamo/utils.py
```diff
@@ -197,3 +201,43 @@ def deparametetrize(
     for key, value in params.names.items():
         expression = expression.replace(key, value)
     return expression
+
+
+@dataclass(frozen=True)
+class MinLen2AppendOnlyList(Generic[T]):
+    first: T
+    second: T
```
dimaqq, line 209: `first` and `second` are not used outside of this dataclass.

How about the following instead:
```py
@dataclass
class Foo:
    data: tuple

    @property
    def first(self): return self.data[0]

    @property
    def second(self): return self.data[1]
```

src/aiodynamo/utils.py
```diff
@@ -197,3 +201,43 @@ def deparametetrize(
...
+        return cls(first, second, rest)
+
+    def appending(self, value: T) -> MinLen2AppendOnlyList[T]:
+        return MinLen2AppendOnlyList(self.first, self.second, (*self.rest, value))
+
+    def extending(self, values: Iterable[T]) -> MinLen2AppendOnlyList[T]:
+        return MinLen2AppendOnlyList(self.first, self.second, (*self.rest, *values))
+
+    def __contains__(self, item: Any) -> bool:
+        return item == self.first or item == self.second or item in self.rest
```
dimaqq, line 223: Minor (probably goes away if refactored):
https://docs.python.org/3/reference/expressions.html#membership-test-operations
`any(x is e or x == e for e in y)`

dimaqq (commented):
Overall thoughts, sorry if imprecise.

I feel that the only use for the new class is the `isinstance` test.
Everything else is a regular tuple.
Would it be better to subclass a plain `tuple` instead?

dimaqq:
> Overall thoughts, sorry if imprecise.
> 
> I feel that the only use for the new class is the `isinstance` test. Everything else is a regular tuple. Would it be better to subclass a plain `tuple` instead?

And I managed to get it all wrong.
Would it be better to change `And/OrCondition`'s `lhs/rhs` to `children` which is a tuple (or a list)?

src/aiodynamo/utils.py
ojii: they're there because this _must_ have at least two elements. it's illegal to create an instance with zero or one elements. that's why `first` and `second` are outside the tuple. they're implicitly used via `__iter__` [here](https://github.com/hennge/aiodynamo/blob/ce6e63f07da562c9a646d3a4c9a35512d8a80d6d/src/aiodynamo/expressions.py#L424) and [here](https://github.com/hennge/aiodynamo/blob/ce6e63f07da562c9a646d3a4c9a35512d8a80d6d/src/aiodynamo/expressions.py#L432)

src/aiodynamo/utils.py
ojii: honestly, most of these magic methods could (should?) probably removed. I just made it as 'sequene-like' as possible for no real reason 😅

src/aiodynamo/utils.py
dimaqq: That can be solved by an assert or a ValueError in the `encode` method, can it not?

ojii: it could at runtime, the goal here was to ensure it statically.

dimaqq:
I've tried to express my thoughts on refactoring in #174, pull it in if you deem it worthy 🙇🏻

src/aiodynamo/utils.py
```diff
@@ -197,3 +201,43 @@ def deparametetrize(
     for key, value in params.names.items():
         expression = expression.replace(key, value)
     return expression
+
+
+@dataclass(frozen=True)
+class MinLen2AppendOnlyList(Generic[T]):
```
y26805, line 207: ideas:
`MinTwoItemAppendOnlyList`
`MinTwoAppendOnlyList`
`AtLeastTwoAppendOnlyList`
`TwoOrMoreAppendOnlyList`
...

dimaqq: How about "Multiple" or "Several"?

On another note, this is more of an immutable list, calling it "append only" feels misleading, as the stored list is never modified after creation.

So maybe a tuple then?

How about "multivariate tuple" which is a thing in statistics, a tuple containing 2 or more elements.

ojii: yea it's no longer a list. it's more of a tuple-like iterable with a minimum length

ojii:
`MinLen2AppendOnlyList` needlessly generic, renamed to `SubConditions` and de-genericized.

PR #174, opened 2024-01-05
Simplify/refactor #173

dimaqq:


dimaqq:
@ojii here

ojii:
sorry, I really prefer ensuring correctness at the type level if possible, so I'm going to close this.

src/aiodynamo/expressions.py
```diff
@@ -377,20 +377,24 @@ def encode(self, params: Parameters) -> str:
...
-            if isinstance(other, AndCondition):
-                return AndCondition(self.children.extending(other.children))
-            else:
-                return AndCondition(self.children.appending(other))
-        return AndCondition(MinLen2AppendOnlyList.create(self, other))
+        if isinstance(self, AndCondition) and isinstance(other, AndCondition):
+            return AndCondition(self.children + other.children)
+        elif isinstance(self, AndCondition):
+            return AndCondition(self.children + (other,))
+        elif isinstance(other, AndCondition):
```
dimaqq, line 384: New and improved: #173 missed this case:
`A and (B and C)` 
That's when the user manually adds parens in non-intuitive way.

ojii: nice, I'll add that to the other PR

dimaqq:
No problem. Consider including the missed logical branch from the comment above.

dimaqq:
> sorry, I really prefer ensuring correctness at the type level if possible, so I'm going to close this.

I was going to suggest a type guard, but that would require py 3.10 ¯\_(ツ)_/¯
https://devguide.python.org/versions/

dimaqq:
Just a thought: another option would be to override AndCondition and OrCondition constructors to ensure the contents is of the correct length.

PR #180, opened 2024-02-16
Separate request and response logging into other named loggers.

aclemons:
I'd like to make the logging for aiodynamo louder at times for debugging
or to generally see in application logs when a throttle or a timeout or
something happened. For that, I need to set the log level to debug, but
this also enables request and response payload logging which becomes
very noisy and also smears potentially unwanted data into my log
streams.

I made two other loggers so the levels for those two things can be
set at a different level to all the other logging done by aiodynamo.

### DIFF
--- a/src/aiodynamo/client.py
+++ b/src/aiodynamo/client.py
@@ -70,7 +70,7 @@ from .models import (
 )
 from .sign import signed_dynamo_request
 from .types import Item, NumericTypeConverter, TableName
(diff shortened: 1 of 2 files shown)


aclemons:
Thoughts on this?

dimaqq:
👉🏻 @ojii

ojii:
thank you

PR #183, opened 2024-05-25
Fix code indent in readme.md

aisk:


dimaqq:
I think that original indentation was intended. 

The example tries to show how to use the library in an asynchronous function and what imports are needed.

After all, you can’t have `async with` at the module level, can you?

aisk:
I think it's not indented, for example with this code in README:

```py
from aiodynamo.client import Client
from aiodynamo.credentials import Credentials
from aiodynamo.http.httpx import HTTPX
from httpx import AsyncClient

    async with AsyncClient() as h:
        client = Client(HTTPX(h), Credentials.auto(), "us-east-1")
```

Run it with Python will got this error:

```
$ python3 a.py
  File "/private/tmp/a.py", line 6
    async with AsyncClient() as h:
IndentationError: unexpected indent
```

There is an extra indent start with line 6.

dimaqq:
Maybe @ojii can make a call on this one.

Doc source code-wise it's fine.

readme.md
```diff
@@ -38,31 +39,33 @@ from aiodynamo.credentials import Credentials
...
-
-        # Create or override an item
-        await table.put_item({"key": "my-item", "value": 1})
-        # Get an item
-        item = await table.get_item({"key": "my-item"})
-        print(item)
-        # Update an item, if it exists.
-        await table.update_item(
-            {"key": "my-item"}, F("value").add(1), condition=F("key").exists()
+async def main():
```
ojii, line 50: `client` below is not defined (it's kind of implicitly supposed to be re-used from the examples above), so could you change this function to take a `client: Client` as argument?

ojii:
reading our current README I'm not sure what I was thinking with that indentation without having stuff in functions, so this proposed patch makes things a lot nicer, thank you @aisk.

readme.md
aisk: Sure, done.

ojii:
thank you

PR #184, opened 2024-05-29
fix a type mismatch for TransactionCanceled

aisk:
The `cancellation_reasons` have a mismatched type when initializing, and this change fixes it. Additionally, it has been changed to use `Optional` instead of `Union[T, None]` because they are equivalent but `Optional` has a shorter name.


ojii (approved):
thank you

PR #185, opened 2024-05-29
add httpx and aiohttp as dev dependencies

aisk:
Because these two libs are required to run the test cases, making it (a tiny bit) easier to set up a new development environment.


dimaqq (approved):
Great idea!

ojii (approved):

ojii:
not a fan of the duplicate dependency declaration needed for this but that seems to be a poetry limitation and the UX/DX of this is worth it in my opinion, thank you.

Issue #70, opened 2021-03-02
Wrap more error responses


dimaqq:
For example, `ResourceInUseException` (e.g. trying to create a table that already exists)

Today the error is `aiodynamo.errors.UnknownError: b'{"__type":"com.amazonaws.dynamodb.v20120810#ResourceInUseException","message":""}'`

flpStrri:
@dimaqq, do you have a list of errors that we should wrap here? I can help with this one.

dimaqq:
Awesome, João!

This particular error we've observed in production, and a bunch are already handled.
I don't know if there's an authoritative list of all possible errors that ddb can throw at us 🙈 

`dynalite` implements [most of] dynamodb protocol, and since it's open source, we should probably support all the errors it produces: https://github.com/mhart/dynalite/search?q=__type

Official list is here: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html

flpStrri:
Sure! I will check both of this resources and put my fingers to work. Also, how small are the PRs here? I should push one by PR, one PR with each error in a commit, pack them all?

dimaqq:
all in one is just fine :)

PR #186, opened 2024-05-30
add more error types

aisk:
Add more pre-defined error types from the [official document](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html).  
  
Also, I added a `SerializationException`, which is not mentioned in the document but is found in dynalite. Someone on Stack Overflow also encountered this error when using DynamoDB, so I assume it is being used by DynamoDB as well.  
  
Related to #70


ojii (approved):
thank you

PR #191, opened 2024-08-15
manually document enums

### LINKED ISSUE #190: autodocs for enums generate crazy output
see https://aiodynamo.readthedocs.io/en/latest/usage.html#aiodynamo.models.ReturnValues

ojii:
Fixes #190

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -387,24 +387,49 @@ Models
     :members: enabled, view_type
     :undoc-members:
 
-.. autoclass:: aiodynamo.models..StreamViewType
-    :members: keys_only, new_image, old_image, new_and_old_images
-    :undoc-members:
+.. py:class:: aiodynamo.models.StreamViewType
+
+    .. py:attribute:: keys_only
+    .. py:attribute:: new_image
+    .. py:attribute:: old_image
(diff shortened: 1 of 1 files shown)


dimaqq:
WDYT about this idiom for documenting enum values?

https://github.com/canonical/operator/blob/6a17bf933114959a968fe65502f429a0dab4452c/ops/charm.py#L1815-L1822

ojii:
those docstrings look like they're the wrong way around (I'd expect them to be before the thing they document)

dimaqq:
The convention was counter-intuitive for me too.

The way I force myself to remember this is:
- after `def foo()` goes the function docstring
- after `attr: int` goes the attribute docstring

dimaqq:
Or you could use https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.Doc if the tooling groks that.

ojii:
> Or you could use https://typing-extensions.readthedocs.io/en/latest/index.html#typing_extensions.Doc if the tooling groks that.

I detest that use of type hints... I do a lot of fastapi these days and I find their doc-typehints to make the code super unreadable.

PR #194, opened 2024-10-15
add support for Process Credentials Provider

ojii:
Adds support for [Process Credentials Provider](https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html)

### DIFF
--- a/src/aiodynamo/credentials.py
+++ b/src/aiodynamo/credentials.py
@@ -572,6 +572,68 @@ class InstanceMetadataCredentialsV1(MetadataCredentials):
         )
 
 
+@dataclass
+class ProcessCredentialsError(Exception):
+    reason: str
+    return_code: int | None
+    stdout: bytes
+    stderr: bytes
+
+
+@dataclass
+class ProcessCredentials(MetadataCredentials):
+    command: list[str]
+
+    async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
+        process = await asyncio.create_subprocess_exec(
+            *self.command,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            stdin=asyncio.subprocess.PIPE,
+        )
+        stdout, stderr = await process.communicate(b"")
+        if process.returncode != 0:
+            raise ProcessCredentialsError(
+                f"Process terminated with non-zero return code {process.returncode}: {try_decode(stderr)}",
+                process.returncode,
+                stdout,
+                stderr,
+            )
+        try:
+            data = json.loads(stdout)
+        except json.JSONDecodeError:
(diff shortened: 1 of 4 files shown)


src/aiodynamo/credentials.py
```diff
@@ -572,6 +572,68 @@ async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
...
+    command: list[str]
+
+    async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
+        process = await asyncio.create_subprocess_exec(
+            *self.command,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            stdin=asyncio.subprocess.PIPE,
+        )
+        stdout, stderr = await process.communicate(b"")
```
dimaqq, line 594: This would block if the subprocess produces too much data.

Arguably there should be a default timeout too, although, perhaps in this particular library, the caller has a timeout and will cancel `fetch_metadata`.

A simple fix could look smth like:

```py
try:
    stdout, stderr = await asyncio.wait_for(
        process.communicate(b""), timeout=30
    )
except asyncio.TimeoutError:
    process.kill()
    await process.wait()
    raise ProcessCredentialsError("Subprocess timed out", None, b"", b"")
```

I dunno if it's worth it in practice.

src/aiodynamo/credentials.py
```diff
@@ -572,6 +572,68 @@ async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
...
+        if process.returncode != 0:
+            raise ProcessCredentialsError(
+                f"Process terminated with non-zero return code {process.returncode}: {try_decode(stderr)}",
+                process.returncode,
+                stdout,
+                stderr,
+            )
+        try:
+            data = json.loads(stdout)
+        except json.JSONDecodeError:
```
dimaqq, line 604: Hmmm I can't recall if they fixed this or not. Back in the day, `except ValueError` was better, because JSONDecodeError inherits from it, and in some special cases loads could raise a plain ValueError?

src/aiodynamo/credentials.py
```diff
@@ -572,6 +572,68 @@ async def fetch_metadata(self, http: HttpImplementation) -> Metadata:
...
+        return Metadata(key, parse_amazon_timestamp(data["expiration"]))
+
+    def is_disabled(self) -> bool:
+        return False
+
+
+def try_decode(data: bytes) -> str:
+    try:
+        return data.decode()
+    except:
```
dimaqq, line 633:
```suggestion
    except Exception:
```

unless the intention is to swallow e.g. SystemExit if it happens at just the right time.

dimaqq (approved):
It would be great to add some super-simple snippet about how to use this in real life.

PR #203, opened 2026-02-19
Upgrade tooling

ojii:
poetry -> uv
isort -> ruff
black -> ruff format

Also, scylla now supports stored throughput configurations and mypy understands that os.environ.get(k, "str") is a str

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,66 +1,46 @@
-[tool.poetry]
+[project]
 name = "aiodynamo"
 version = "24.7"
 description = "Asyncio DynamoDB client"
-authors = ["Jonas Obrist <jonas.obrist@hennge.com>"]
-license = "Apache-2.0"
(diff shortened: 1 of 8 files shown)


jgaul3 (approved):

KazuyaHayashi (approved):

pyproject.toml
```diff
@@ -1,83 +1,63 @@
...
 select = [
-    "E", "F", "T201",
+    "E", "F", "T201", "I"
 ]
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
dimaqq, line 63: Minor: we're considering moving from hatchling to `uv-build` in our open-source projects since we're using uv anyway.

PR #205, opened 2026-02-25
Support dual-stack endpoints through configuration.

aclemons:
AWS recently announced dual-stack endpoints for dynamodb:

https://aws.amazon.com/about-aws/whats-new/2025/10/amazon-dynamodb-ipv6/

Currently, we need to manually construct the url and pass a YARL object
into aiodynamo's constructor as the `endpoint` arg:

```py
endpoint = URL.build(
    scheme="https", host=f"dynamodb.{region}.api.aws", path="/"
)

client = Client(
    AIOHTTP(client_session),
    region=region,
    credentials=Credentials.auto(),
    endpoint=endpoint,
)
```

It would be convenient if aiodynamo supported a `dual_stack` flag like boto does.


src/aiodynamo/sign.py
```diff
@@ -18,7 +18,10 @@
 ALGORITHM = "AWS4-HMAC-SHA256"
 
 
-def make_default_endpoint(region: str) -> URL:
+def make_default_endpoint(region: str, dual_stack: bool) -> URL:
+    if dual_stack:
+        return URL.build(scheme="https", host=f"{SERVICE}.{region}.api.aws", path="/")
```
dimaqq, line 23: Looking through https://docs.aws.amazon.com/general/latest/gr/ddb.html there's a whole bunch of regions that don't support dual-stack endpoints today, even though some months ago they promised that most regions were coming within weeks.

This, coupled with the code here, means that a person configuring an application that uses aiodynamo has to be more careful than before.

At the same time, perhaps that page has not been updated?

```command
> curl https://dynamodb.us-east-2.api.aws
healthy: dynamodb.us-east-2.amazonaws.com ⏎
> curl https://dynamodb.ap-northeast-1.api.aws
healthy: dynamodb.ap-northeast-1.amazonaws.com ⏎
```

While the latter hostname is not mentioned in the page.

Does boto use the same logic for inferring the hostname?

dimaqq:
Also note https://docs.aws.amazon.com/sdkref/latest/guide/feature-account-endpoints.html

Which implies that the endpoint hostname computation used in this library should be updated to match.

(I imagine that would be a separate PR)

src/aiodynamo/sign.py
aclemons: I left the default as `False`, so I don't think you have to be more careful?

But yes, Tokyo for example only had support enabled in January where as some regions were available immediately after the announcement.

aclemons: We had been waiting and finally enabled it with code in the PR description a few weeks back.

dimaqq: I'll let the maintainer decide. I'm not one.

ojii:
I like supporting this, though I'm not a fan of the API for this, because passing `endpoint=some_url, dual_stack=True` is meaningless. I would prefer updating `endpoint` from `None | URL|` to `None | URL | Callable[[str], URL]` (the passed str being the region) and having a constant `DUAL_STACK = lambda region: URL.build(scheme="https", host=f"{SERVICE}.{region}.api.aws", path="/")` (probably an actual function not a lambda but you get the point). Then you'd make a client: `Client(..., endpoint=DUALSTACK)`. 

Also, this should be documented and a note should be added regarding dualstack not being available in all regions.

aclemons:
I can make that change. When I opened this, I was handling it like boto3 does where you can pass in `use_dualstack_endpoint=True` but also an `endpoint_url` which as you pointed out is nonsense. But we don't have to mirror how boto does it - I'll update and push up a change for you to check.

PR #206, opened 2026-03-06
Support multi-attribute GSI partition and sort keys

### LINKED ISSUE #202: Support DynamoDB's new multi-attribute GSI partition and sort keys
DynamoDB added a new feature late last year to support multi-attribute GSI partition and sort keys. It would be great if this library could be updated to support this new feature.

https://aws.amazon.com/about-aws/whats-new/2025/11/amazon-dynamodb-multi-attribute-composite-keys-global-secondary-indexes/
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.DesignPattern.MultiAttributeKeys.html

cluebbehusen:
Closes #202

DynamoDB added support for multi-attribute composite keys on GSIs late last year. This PR adds support for that feature.

Docs: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.DesignPattern.MultiAttributeKeys.html

**KeySchema** now accepts tuples of `KeySpec` for `hash_key` and `range_key`, with validation enforcing DynamoDB's 1-4 attribute limit per key type. Single `KeySpec` values still work as before.

**MultiHashKey** is a new expression class for querying multi-attribute partition keys:

```python
# Table creation
gsi = GlobalSecondaryIndex(
    name="tenant-region-index",
    schema=KeySchema(
        hash_key=(KeySpec("tenant", KeyType.string), KeySpec("region", KeyType.string)),
        range_key=(KeySpec("date", KeyType.string), KeySpec("seq", KeyType.number)),
    ),
    projection=Projection(ProjectionType.all),
    throughput=None,
)

# Querying
async for item in client.query(
    "my-table",
    MultiHashKey(("tenant", "acme"), ("region", "us-east"))
    & RangeKey("date").equals("2025-01-01")
    & RangeKey("seq").gt(0),
    index="tenant-region-index",
):
    pass
```

Also fixes `HashAndRangeKeyCondition` missing `__and__`, which meant chaining multiple `RangeKey` conditions required explicit parentheses.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -34,8 +34,8 @@ jobs:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         implementation: [
           {flavor: "dynalite", name: "dynalite"},
-          {flavor: "other", name: "dynamodb-local"},
-          {flavor: "other", name: "localstack"},
+          {flavor: "dynamodb-local", name: "dynamodb-local"},
+          {flavor: "localstack", name: "localstack"},
           {flavor: "scylla", name: "scylla"}
         ]
     runs-on: "ubuntu-latest"
--- a/src/aiodynamo/expressions.py
+++ b/src/aiodynamo/expressions.py
@@ -241,6 +241,9 @@ class KeyCondition(metaclass=abc.ABCMeta):
     def encode(self, params: Parameters) -> str:
         pass
 
+    def __and__(self, other: Condition) -> KeyCondition:
+        raise NotImplementedError
+
 
 @dataclass(frozen=True)
 class HashKey(KeyCondition):
@@ -375,6 +378,9 @@ class HashAndRangeKeyCondition(KeyCondition):
 
     def encode(self, params: Parameters) -> str:
         return f"{self.hash_key.encode(params)} AND {self.range_key_condition.encode(params)}"
+
+    def __and__(self, other: Condition) -> KeyCondition:
+        return HashAndRangeKeyCondition(self.hash_key, self.range_key_condition & other)
 
 
 class Condition(metaclass=abc.ABCMeta):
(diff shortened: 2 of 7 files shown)


dimaqq (commented):
I'm not the maintainer, but used to work on this codebase.

I think a change like this should come with a new or updated integration test.

cluebbehusen:
> I'm not the maintainer, but used to work on this codebase.
> 
> I think a change like this should come with a new or updated integration test.

Thanks, that's a good point! I've added integration tests in the latest commit. They cover various cases that DynamoDB supports.

Unfortunately, not all integration testing backends support this new feature yet. DynamoDB local does, and I validated the tests against that backend. Unfortunately, Scylla and dynalite do not, so I set up the tests to skip on those backends.

FWIW, [moto](https://github.com/getmoto/moto) supports this new feature, but I didn't see that as an existing backend. I'd be happy to add that though if desired.

ojii:
> Unfortunately, not all integration testing backends support this new feature yet.

you can use the [`flavor`](https://github.com/HENNGE/aiodynamo/blob/952d0a7a9f186ad5d762234d22457cfb689c1399/tests/integration/conftest.py#L97-L99) fixture to skip tests on some implementations.

cluebbehusen:
> > Unfortunately, not all integration testing backends support this new feature yet.
> 
> you can use the [`flavor`](https://github.com/HENNGE/aiodynamo/blob/952d0a7a9f186ad5d762234d22457cfb689c1399/tests/integration/conftest.py#L97-L99) fixture to skip tests on some implementations.

@ojii, I did set up the tests to skip on unsupported backends.

One nuance is that I had to base it off of the backend name rather than `flavor`. I did this because there is no `flavor` for DynamoDB local or localstack; they are both `other`. DynamoDB local works, but localstack does not, so I could not control skipping based on `other`.

I would be happy to add more granular `flavor`s, but that felt like scope creep on this PR. Let me know if you would prefer expanding `flavor` though, and I'd be happy to make that change.

ojii:
> I would be happy to add more granular `flavor`s, but that felt like scope creep on this PR. Let me know if you would prefer expanding `flavor` though, and I'd be happy to make that change.

expanding makes sense. basically I only added different flavors if they were needed because the implementations differed (I think currently we don't use it since all implementations are similar enough for the tests)

cluebbehusen:
> > I would be happy to add more granular `flavor`s, but that felt like scope creep on this PR. Let me know if you would prefer expanding `flavor` though, and I'd be happy to make that change.
> 
> expanding makes sense. basically I only added different flavors if they were needed because the implementations differed (I think currently we don't use it since all implementations are similar enough for the tests)

Perfect! I expanded `flavor` and updated the tests to use that. I removed the `implementation_name` fixture I had added since it's now unused.

cluebbehusen:
@ojii I resolved conflicts on this PR

PR #208, opened 2026-04-08
feat(trace): add hooks for telemetry

### LINKED ISSUE #207: add xray/opentelemetry support
would be nice. not entirely sure what is needed for this though

kasteph:
## Related Issue
Closes #207 

## Description
This PR adds telemetry hooks to `aiodynamo`.

Using telemetry with `aiodynamo.client.Client` is optional. When omitted, the client uses the default no-op tracer and meter implementations.

The telemetry APIs follow the shape of [`opentelemetry.trace`](https://opentelemetry-python.readthedocs.io/en/latest/_modules/opentelemetry/trace.html) and [`opentelemetry.metrics`](https://opentelemetry-python.readthedocs.io/en/latest/_modules/opentelemetry/metrics/_internal.html). This means that OpenTelemetry-compatible backends can also be used.

### Features

- Add the `aiodynamo.otel` module containing types for integration with `OpenTelemetry` and a convenience object, `metrics.ClientMetrics`, to simplify requests and batch metrics from `aiodynamo.client.Client`.
- Add OpenTelemetry as an optional dependency for `aiodynamo`.
- Added unit tests for types, metric helper behavior, and integration coverage for batch instrumentation.

### Documentation Changes

- Added documentation in `docs/` for information on how to use the `Telemetry` object with `aiodynamo.client.Client`.

### Requisite Services
Run the following Docker containers:
- DynamoDB - whether it's `amazon/dynamodb-local` or `dynalite`

### Minimal Script

Install `opentelemetry` from the project with `uv sync --extra opentelemetry`. The script below will output the traces and metrics to stdout.

```py3
import asyncio
import os

import aiohttp
from opentelemetry import trace
from opentelemetry.sdk.metrics import MeterProvider
from opentelemetry.sdk.metrics.export import ConsoleMetricExporter, PeriodicExportingMetricReader
from opentelemetry.sdk.resources import Resource
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import ConsoleSpanExporter, SimpleSpanProcessor
from yarl import URL

from aiodynamo.client import Client
from aiodynamo.credentials import EnvironmentCredentials
from aiodynamo.http.aiohttp import AIOHTTP
from aiodynamo.otel.types import Telemetry


def init_otel():
    resource = Resource.create({"service.name": "aiodynamo-local-test"})

    tracer_provider = TracerProvider(resource=resource)
    tracer_provider.add_span_processor(SimpleSpanProcessor(ConsoleSpanExporter()))
    trace.set_tracer_provider(tracer_provider)

    metric_reader = PeriodicExportingMetricReader(ConsoleMetricExporter())
    meter_provider = MeterProvider(resource=resource, metric_readers=[metric_reader])

    return tracer_provider, meter_provider


async def main():
    tracer_provider, meter_provider = init_otel()
    session = aiohttp.ClientSession()

    try:
        os.environ["AWS_ACCESS_KEY_ID"] = "test"
        os.environ["AWS_SECRET_ACCESS_KEY"] = "foo"

        telemetry = Telemetry(
            tracer=trace.get_tracer(__name__),
            meter=meter_provider.get_meter(__name__),
        )

        client = Client(
            http=AIOHTTP(session),
            credentials=EnvironmentCredentials(),
            region="us-east-1",
            endpoint=URL("http://localhost:8000"),
            telemetry=telemetry,
        )

        table_name = "otel_test_table"
        item = {"pk": "user#1", "sk": "profile", "name": "Alice"}

        await client.put_item(table_name, item)
        result = await client.get_item(table_name, {"pk": "user#1", "sk": "profile"})
        print(result)

    finally:
        await session.close()
        tracer_provider.shutdown()
        meter_provider.shutdown()


if __name__ == "__main__":
    asyncio.run(main())
```

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -7,7 +7,7 @@ Client instantiation
 You should try to re-use clients, and especially HTTP clients, as much as possible. Don't create a new one for each
 action.
 
-The :py:class:`aiodynamo.client.Client` class takes three required and three optional arguments:
+The :py:class:`aiodynamo.client.Client` class takes three required and four optional arguments:
 
 1. An HTTP client adaptor, conforming to the :py:class:`aiodynamo.http.base.HTTP` interface.
 2. An instance of :py:class:`aiodynamo.credentials.Credentials` to authenticate with DynamoDB. You may use
@@ -16,6 +16,10 @@ The :py:class:`aiodynamo.client.Client` class takes three required and three opt
 4. An optional endpoint URL of your DynamoDB, as a :py:class:`yarl.URL` instance. Useful when using a local DynamoDB implementation such as dynalite or dynamodb-local.
 5. Which numeric type to use. This should be a callable which accepts a string as input and returns your numeric type as output. Defaults to ``float``.
 6. The throttling configuration to use. An instance of :py:class:`aiodynamo.models.RetryConfig`. By default, if the DynamoDB rate limit is exceeded, aiodynamo will retry up for up to one minute with increasing delays.
+7. Optional telemetry provider, used to emit traces and metrics. An instance of :py:class:`aiodynamo.otel.types.Telemetry`. Supply an OpenTelemetry-backed implementation or it defaults to no-op telemetry.
+
+.. seealso::
+    :ref:`telemetry` for more details.
 
 Credentials
 -----------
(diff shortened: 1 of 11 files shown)


dimaqq:
This is becoming a very large pull request.

My vote is a No.

kasteph:
I’m happy to break this down. Given #207 is an open goal, what’s the preferred approach here? If the hooks are too much, I can move to a simpler lazy-import pattern, or just split the current implementation into smaller PRs to make it easier to digest.

ojii:
I have no experience with otel or similar stuff and have not yet in detail read the PR, but it does seem quite large and invasive to me. However, this does seem like a very useful feature (and something that ...boto... libraries support out the box as far as i know), so I'd like to include it.

Before I dive deeper into this, I'm curious why this cannot be as simple as (this simplified pseudocode):

```python
class Tracer(Protocol):
    def attempt(self) -> None:
        ...

    def exception(self, exc: Exception) -> None:
        ...

    def success(self, response: dict[str, Any]) -> None:
        ...


class NullTracer:
    def attempt(self) -> None:
        pass

    def exception(self, exc: Exception) -> None:
        pass

    def success(self, response: dict[str, Any]) -> None:
        pass

@contextmanager
def no_tracing(_, __) -> Generator[NullTracer]:
    yield NullTracer()

type Action = str
type Payload = dict[str, Any]

type Tracing = Callable[[Action, Payload], ContextManager[Tracer]]


class Client:
    tracing: Tracing = no_tracing
    async def send_request(self, action: str, payload: dict[str, Any]) -> dict[str, Any]:
        with self.tracing(action, payload) as tracer:
            async for _ in self.throttle_config.attempts():
                tracer.attempt()
                try:
                    response = self.http(Request(...))
                except Exception as exc:
                    tracer.exception(exc)
                else:
                    data = json.loads(response)
                    tracer.success(data)
                    return data
```

Specifically, why do we need to add telemetry related code to so many methods?

PR #212, opened 2026-06-09
Support EKS container credentials (EKS Pod Identity)

FurqanHabibi:
Reference : https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html

The main difference between EKS and ECS container credentials is that EKS provides `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` while ECS provides `AWS_CONTAINER_AUTHORIZATION_TOKEN`. Other than that EKS also provides `AWS_CONTAINER_CREDENTIALS_FULL_URI` which is already covered by existing code.

I haven't added any test. It seems like the existing ECS container credentials doesn't have test either. Let me know if I should add a test.

Note that I've tested this and it's currently being used on my own EKS cluster.

### DIFF
--- a/src/aiodynamo/credentials.py
+++ b/src/aiodynamo/credentials.py
@@ -343,11 +343,11 @@ def and_then(thing: Optional[T], mapper: Callable[[T], U]) -> Optional[U]:
     return thing
 
 
-# https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint.html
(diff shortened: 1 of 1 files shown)


lujamitch (approved):
LGTM 
I'm not sure if it's okay to merge this straight into `master` though. Maybe it's fine 🤔

hanscg (approved):

PR #215, opened 2026-07-21
Add exponential backoff with jitter to metadata credential retries

### LINKED ISSUE #193: ContainerMetadataCredentials should retry with backoff
When running aiodynamo in an ECS application under high load it is possible to hit timeout issues accessing the container metadata endpoint. I believe the chance of re-connection could be improved if the retry logic implemented some [basic backoff](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/operating-at-scale-dealing-with-throttles.html).

```
Traceback (most recent call last):
  File "uvloop/cbhandles.pyx", line 63, in uvloop.loop.Handle._run
  File "/opt/venv/lib/python3.12/site-packages/aiodynamo/credentials.py", line 289, in _set_current
    self._current = task.result()
                    ^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/aiodynamo/credentials.py", line 389, in fetch_metadata
    response = await fetch_with_retry_and_timeout(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/aiodynamo/credentials.py", line 604, in fetch_with_retry_and_timeout
    raise TooManyRetries()
aiodynamo.credentials.TooManyRetries
```

medhaholla7:
Fixes #193

`fetch_with_retry_and_timeout` retried immediately with no delay between
attempts, meaning `ContainerMetadataCredentials` (default max_attempts=3)
would hammer the ECS metadata endpoint back-to-back on transient failures
instead of backing off.

This adds exponential backoff with jitter (base_delay=0.1s, max_delay=5s,
both configurable) between retry attempts, skipping the delay before the
final attempt. Jitter is included so multiple containers retrying
simultaneously don't all retry in lockstep.

Note: this function is shared by InstanceMetadataCredentialsV2 and
InstanceMetadataCredentialsV1 too, but both default to max_attempts=1,
so they're unaffected unless someone raises that value.

Added three tests:
- confirms the delay actually happens and grows between attempts
- confirms the original exception still propagates after exhausting all attempts
- confirms TooManyRetries is raised correctly when every attempt times out
  (this path had no prior test coverage)

### DIFF
--- a/src/aiodynamo/credentials.py
+++ b/src/aiodynamo/credentials.py
@@ -6,7 +6,6 @@ import configparser
 import datetime
 import json
 import os
-import random
 from dataclasses import dataclass, field
 from enum import Enum, auto
 from pathlib import Path
@@ -15,6 +14,7 @@ from typing import Callable, Coroutine, Generic, Optional, Sequence, TypeVar, Un
 from yarl import URL
 
 from .http.types import HttpImplementation, Request, RequestFailed
+from .models import ExponentialBackoffRetry
 from .types import Timeout
 from .utils import logger, parse_amazon_timestamp
 
@@ -573,10 +573,9 @@ async def fetch_with_retry_and_timeout(
     max_attempts: int,
     timeout: Timeout,
     request: Request,
-    base_delay: float = 0.1,
-    max_delay: float = 5.0,
 ) -> bytes:
     exception: Optional[Exception] = None
+    delays = iter(ExponentialBackoffRetry(max_delay_secs=2).delays())
     for attempt in range(max_attempts):
         try:
             response = await asyncio.wait_for(http(request), timeout)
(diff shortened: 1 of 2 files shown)


ojii:
we already have [`RetryConfig`](https://github.com/HENNGE/aiodynamo/blob/f326b87b4cf0a2106d5c80a1c41b5b858f8aa850/src/aiodynamo/models.py#L283C7-L349) which supports backoff for retries. can this be reused somehow instead of re-implementing it?

medhaholla7:
I swapped it to use ExponentialBackoffRetry from models.py directly instead of hand-rolling it. Also capped max_delay_secs at 2s since this is fast local metadata.

